PR #80, opened 2022-03-25
DVC-2796 update AS build system, implement murmurhash logic

jonathannorris:


lib/shared/bucketing-test-data/src/index.ts
```diff
@@ -0,0 +1,21 @@
...
+}
+
+fs.writeFileSync(path.resolve(__dirname, `${folderName}/testData.json`), JSON.stringify(testData))
+fs.writeFileSync(path.resolve(__dirname, `${folderName}/generatedConfig.json`), JSON.stringify(generatedConfig))
+console.log(`Wrote test data to ${folderPath}`)
+
+export * from './data/testData'
+export * from './data/generatedConfig'
+//
+// import testDataJSON from '../json-data/testData.json'
```
ajwootto, line 19: remove this?

lib/shared/bucketing-test-data/src/data/testData.ts
```diff
@@ -190,34 +190,34 @@ export const variations: PublicVariation[] = [
...
-                _var: variables[3]._id,
-                value: 610
-            },
-            {
-                _var: variables[4]._id,
-                value: 1114
-            }
-        ]
-    }
+    // {
```
ajwootto, line 193: how come these are commented?

lib/shared/bucketing/__test__/bucketing.test.ts
```diff
@@ -7,7 +7,7 @@ import {
     generateBucketedConfig,
     doesUserPassRollout
 } from '../src/bucketing'
-import { config, barrenConfig } from './data/testData'
+import { config, barrenConfig } from '@devcycle/bucketing-test-data/src/data/testData'
```
ajwootto, line 10: is this not able to just import from the index anymore?

lib/shared/bucketing-test-data/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "@devcycle/bucketing-test-data",
+  "version": "1.0.0",
+  "devDependencies": {
```
ajwootto, line 4: probably delete this dependency, its declared in the root

lib/shared/bucketing/__test__/bucketing.test.ts
jonathannorris: no, because that would execute the writing to disk

ajwootto: maybe you could do something like check for a command-line flag to enable the writing, and update the target definition to pass that flag?

jonathannorris: I don't think we can, because that would cause the files to export the TS code with the links to the Types library that the AS code was having issues with. We set up this whole build system so it only directly exports the JSON data.

ajwootto (approved):

PR #82, opened 2022-03-29
add new validator for blank user ids

elliotCamblor:


lib/shared/types/src/types/apis/sdk/clientSDKAPI.ts
```diff
@@ -38,6 +39,7 @@ export class DVCAPIUser {
      * Must be defined if `isAnonymous = false`
```
andrewtraviss, line 39: This comment implies that this field should be optional, for the isAnonymous: true scenario?

andrewtraviss (approved):

PR #83, opened 2022-03-29
DVC-930 add ISO-639-1 validator

squarepegg:


lib/shared/types/src/types/apis/sdk/clientSDKAPI.ts
```diff
@@ -198,8 +200,9 @@ export class DVCClientAPIUser implements DVCAPIUser {
         name?: string
 
     /**
-     * ISO 639-1 two letter codes
+     * ISO 639-1 two-letter codes
      */
+    @IsISO6391()
```
ajwootto, line 205: should we be collecting the language automatically in the SDKs like we did in TL, or leave it up to the user?

squarepegg: A good discussion point, but a far different change than addressing an outstanding issue.

Happy to put this on hold until this we can bring this up at the next sdk guild.

ajwootto (approved):

PR #87, opened 2022-04-04
feat: add helper method to sort array of objects by a provided value

ajwootto:


lib/shared/bucketing-assembly-script/assembly/helpers/arrayHelpers.ts
```diff
@@ -0,0 +1,50 @@
...
+        }
+
+        // return subtraction of the current cursor position's character codes (which must be different)
+        return a.value.charCodeAt(cursor) - b.value.charCodeAt(cursor)
+    })
+
+    let result: T[] = []
+    for (let i = 0; i < sorted.length; i++) {
+        result.push(sorted[i].entry)
+    }
```
jonathannorris, line 47: you could do a `.map()` on this if you want.

lib/shared/bucketing-assembly-script/assembly/index.ts
```diff
@@ -31,3 +32,10 @@ export function testBucketedUserConfigClass(userConfigStr: string): string {
 export function testMurmurhashV3(key: string, seed: i32): i32 {
     return murmurhashV3(key, seed)
 }
+
+class TestData {
+    key: string
+}
+export function testSortObjectsByString(arr: SortingArray<TestData>): TestData[]{
```
jonathannorris, line 39: does calling this like this from JS actually work without JSON parsing?

jonathannorris: interested in what it looks like in: `bucketing-lib.debug.d.ts` / `bucketing-lib.debug.js`

lib/shared/bucketing-assembly-script/assembly/helpers/arrayHelpers.ts
ajwootto: I actually tried that but it refused to compile and complained that the types were wrong, no matter how explicitly I wrote them out (it was fine in Webstorm)

lib/shared/bucketing-assembly-script/assembly/index.ts
ajwootto: It seems to output a function that receives an array, but either way the tests are passing so that must mean it works?

lib/shared/bucketing-assembly-script/assembly/helpers/arrayHelpers.ts
jonathannorris: I think if you set the return type in `.map<T>()` it works: 
image

jonathannorris (approved):

lib/shared/bucketing-assembly-script/assembly/helpers/arrayHelpers.ts
ajwootto: tried that

PR #88, opened 2022-04-05
DVC-2583 setting up bucketing / segmentation tests

jonathannorris:


lib/shared/bucketing-assembly-script/assembly/bucketing/bucketing.ts
```diff
@@ -18,7 +18,15 @@ class BoundedHash {
...
-export function generateBoundedHashes(user_id: string, target_id: string): BoundedHash {
+export function generateBoundedHashesFromJSON(user_id: string, target_id: string): string {
+    const boundedHash = _generateBoundedHashes(user_id, target_id)
+    const json = new JSON.Obj()
+    json.set('rolloutHash', boundedHash.rolloutHash)
+    json.set('bucketingHash', boundedHash.bucketingHash)
+    return json.stringify()
+}
+
+export function _generateBoundedHashes(user_id: string, target_id: string): BoundedHash {
```
ajwootto, line 29: what are the underscore prefixes meant to represent?

lib/shared/bucketing-assembly-script/assembly/bucketing/segmentation.ts
```diff
@@ -2,10 +2,21 @@ import { RegExp } from 'assemblyscript-regex/assembly'
...
 import {
-    TopLevelOperator, AudienceFilterOrOperator, DVCPopulatedUser, validSubTypes
+    TopLevelOperator, AudienceFilterOrOperator, DVCPopulatedUser, validSubTypes,
+    DVCUser, Target as PublicTarget
 } from '../types'
 import { JSON } from 'assemblyscript-json/assembly'
-import { getF64FromJSONValue } from '../helpers/jsonHelpers'
+import {getF64FromJSONObj, getF64FromJSONValue} from '../helpers/jsonHelpers'
+
+export function evaluateOperatorFromJSON(operatorStr: string, userStr: string): bool {
```
ajwootto, line 11: maybe it would be cleaner if you collected all the public interface methods in one file and then just had them call the methods from the existing bucketing and segmentation files? With the current structure it's pretty hard to easily see what the interface of this library is.

lib/shared/bucketing-assembly-script/assembly/index.ts
```diff
@@ -1,17 +1,22 @@
 import {
     BucketedUserConfig, ConfigBody, DVCUser, DVCPopulatedUser
 } from "./types"
-import * as bucketing from './bucketing'
+import {
```
ajwootto, line 4: you could also use `export from`

lib/shared/bucketing-test-data/json-data/testData.json
```diff
@@ -1 +1 @@
...
..."user","subType":"customData","dataKey":"favouriteFood","comparator":"=","values":["pizza"]},{"type":"user","subType":"customData","dataKey":"favouriteDrink","comparator":"=","values":["coffee"]}],"operator":"and"}},"distribution":[{"_variation":"615382338424cb11646d7667","percentage":1}]}]},"variations":[{"_id":"6153553b8cf4e45e0464268d","name":"variation 1","variables":[{"_var":"614ef6ea475129459160721a","value":"scat"},{"_var":"615356f120ed334a6054564c","value":"man"}]},{"_id":"615357cf7e9ebdca58446ed0","name":"variation 2","variables":[{"_var":"615356f120ed334a6054564c","value":"YEEEEOWZA"}]},{"_id":"615382338424cb11646d7667","name":"variation 1 aud 2","variables":[{"_var":"61538237b0a70b58ae6af71f","value":"Var 1 aud 2"}]},{"_id":"615382338424cb11646d7668","name":"feature 2 variation","variables":[{"_var":"61538237b0a70b58ae6af71g","value":"multivar first"},{"_var":"61538237b0a70b58ae6af71h","value":"multivar last"}]},{"_id":"615382338424cb11646d7668","name":"feature 2 never used variation","variables":[{"_var":"61538237b0a70b58ae6af71g","value":"multivar first unused"},{"_var":"61538237b0a70b58ae6af71h","value":"multivar last unused"}]}]}],"variables":[],"variableHashes":{}}}
```
ajwootto, line 1: it would be nice if this file was formatted so we could see the diff properly

lib/shared/bucketing-assembly-script/project.json
```diff
@@ -30,10 +30,10 @@
             "options": {
                 "commands": [
                     {
-                      "command": "asc assembly/index.ts --target debug --exportRuntime --initialMemory 100 --maximumMemory 10000 --bindings esm"
+                      "command": "asc assembly/index.ts --target debug --exportRuntime --initialMemory 10000 --maximumMemory 20000 --bindings esm"
```
JamieSinn, line 33: not 50k max?

JamieSinn (approved):

lib/shared/bucketing-assembly-script/assembly/bucketing/bucketing.ts
jonathannorris: internal methods vs external ones.

lib/shared/bucketing-assembly-script/project.json
ajwootto: It seems to work either way, do we need a higher limit? Was that based on the math you did?

JamieSinn: 50k was based on the GH usage math - more futureproofing than anything

lib/shared/bucketing-assembly-script/assembly/bucketing/segmentation.ts
```diff
@@ -85,11 +85,11 @@ export const convertToSemanticVersion = (version: string): string => {
     return splitVersion.join('.')
 }
 
-export const checkVersionValue = (
+export function checkVersionValue(
     filterVersion: string | null,
     version: string | null,
     operator: string | null
-): bool => {
+): bool{
```
jonathannorris, line 92:
```suggestion
): bool {
```

lib/shared/bucketing-assembly-script/project.json
jonathannorris, line 33: @JamieSinn / @ajwootto Do we still need the initial memory to be this high?

jonathannorris: 50K pages would be 3.2 GB of memory... (each page is 64kb)

ajwootto: yup otherwise the tests won't pass

jonathannorris: can we just set it for the debug build then?

JamieSinn: It'll be required for any sort of use at scale in production too

jonathannorris: but does it need 1.2GB? (20,000 * 64kb pages)

JamieSinn: Yep - this is assuming that each config is 2-3MB - or about 45 pages. Basing it off GH/RBC's per second request load - we can assume about 1200 users would need to be bucketed at any one time - 45 * 1200 = ~50k pages of RAM

JamieSinn: This is the scale issue i was talking about

jonathannorris: but they will all be using the same cached config.

JamieSinn: 20k is the initial load - if that's not used the GC will free them and scale up as needed.

JamieSinn: but also -50k isn't automatically allocated - it's just the max that the GC can allocate total at any one time

PR #101, opened 2022-04-11
feat: add method to set config body in advance

ajwootto:
- add method to set config data so that it is available when bucketed configs are requested
- removed config param from bucketed config method

### DIFF
--- a/lib/shared/bucketing-assembly-script/assembly/managers/configDataManager.ts
+++ b/lib/shared/bucketing-assembly-script/assembly/managers/configDataManager.ts
@@ -1,0 +1,15 @@
+import { ConfigBody } from '../types'
+
+let _configData: Map<string, ConfigBody> = new Map()
+
+export function _setConfigData(token: string, configData: ConfigBody): void {
+    _configData.set(token, configData)
+}
+
+export function _getConfigData(token: string): ConfigBody {
+    if (!_configData.has(token)) {
+        throw new Error('Config data is not set.')
+    } else {
+        return _configData.get(token)
+    }
+}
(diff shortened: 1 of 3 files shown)


lib/shared/bucketing-assembly-script/assembly/managers/configDataManager.ts
```diff
@@ -0,0 +1,15 @@
+import { ConfigBody } from '../types'
+
+let _configData: ConfigBody | null = null
+
+export function _setConfigData(configData: ConfigBody): void {
```
jonathannorris, line 5: How do the existing interfaces for the Server SDKs work? do they only support a single project? Or can you create multiple instances for different projects?

ajwootto: the client itself needs to be instantiated with a single project token, so if you wanted to use multiple projects you'd have to make multiple clients. I guess this wouldn't support that since it's basically a global singleton

ajwootto: so should the signature for the generateBucketedConfig require a token as well then?

jonathannorris: Yea it would need to.

lib/shared/bucketing-assembly-script/assembly/managers/configDataManager.ts
```diff
@@ -0,0 +1,15 @@
+import { ConfigBody } from '../types'
+
+let _configData: Map<string, ConfigBody> = new Map()
```
jonathannorris, line 3: I wonder if we should set a size limit on this somehow? doesn't look like we can do that with a normal Map, would need to do it with an Array, which isn't great.

jonathannorris (approved):

lib/shared/bucketing-assembly-script/assembly/managers/configDataManager.ts
ajwootto: I was thinking we didn't need to because it would be up to the people using the server SDK to not use too many, but I just remembered we also want to use this for CF workers.

PR #102, opened 2022-04-12
Escape non-ascii characters before murmurhash function

jonathannorris:
- also refactored tests into their own sub-folders


JamieSinn (approved):

lib/shared/bucketing-assembly-script/assembly/types/dvcUser.ts
```diff
@@ -16,6 +16,16 @@ interface DVCUserInterface {
     privateCustomData: JSON.Obj | null
 }
 
+export function stringIsAscii(str: string): boolean {
```
ajwootto, line 19: is this a duplicate?

ajwootto (approved):

PR #103, opened 2022-04-12
feat: add React Native example app

chantalchung:
- added the `browser` field in `package.json` for the react, js sdk so that the package resolves properly for the React Native app
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
     "@nrwl/next": "13.9.3",
     "@nrwl/node": "13.9.3",
     "@nrwl/react": "13.9.3",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -67,8 +71,13 @@
     "@nrwl/next": "13.9.3",
     "@nrwl/node": "13.9.3",
     "@nrwl/react": "13.9.3",
+    "@nrwl/react-native": "^13.10.1",
```
ajwootto, line 74: make sure this is the same version as all the other `@nrwl` packages

jsalaber (approved):
#

PR #105, opened 2022-04-12
chore: add linting and fix linting problems

ajwootto:


lib/shared/bucketing-test-data/.gitignore
```diff
@@ -0,0 +1 @@
+json-data/*
```
ajwootto, line 1: removing the json data from git because it has dates in it that are recomputed on every run and cause tests to fail if they aren't relative to the current date

nx.json
```diff
@@ -23,8 +23,7 @@
           "lint",
           "check-types",
           "package",
-          "prepare",
-          "build:json"
```
ajwootto, line 27: caching this was causing test failures due to dates

jonathannorris (approved):

PR #107, opened 2022-04-12
chore: switch custom data filter evaluation to use dataKeyType

ajwootto:


lib/shared/bucketing-assembly-script/assembly/types/target.ts
```diff
@@ -172,6 +171,24 @@ export class AudienceFilterOrOperator extends JSON.Value {
...
 }
 
+export class CustomDataFilter extends AudienceFilterOrOperator {
+    dataKeyType: string
+    dataKey: string
+
+    constructor(filter: JSON.Obj) {
+        super(filter)
+        this.dataKey = getStringFromJSON(filter, 'dataKey')
+        this.dataKeyType = isValidString(filter, 'dataKeyType', validDataKeyTypes)
```
jonathannorris, line 181: how is this different than the constructor above? 
image

jonathannorris: ahh so it makes it non-optional?

jonathannorris (approved):

lib/shared/bucketing-assembly-script/assembly/types/target.ts
ajwootto: yeah, its basically just for input data validation because it still gets passed around as an AudienceOrOperator object. I'm then explicitly casting it to the CustomDataFilter when it gets used as that, which isn't great. Can't find a better way to do it though due to the lack of union types

PR #111, opened 2022-04-13
fix: change identify and reset to be able to be called before client is initialized

jsalaber:
# Changes

- change identify and reset to handle config being undefined at the time of call to support client being called before config comes

### DIFF
--- a/sdk/js/src/Client.ts
+++ b/sdk/js/src/Client.ts
@@ -110,13 +110,6 @@ export class DVCClient implements Client {
         user: DVCUser,
         callback?: ErrorCallback<DVCVariableSet>
     ): Promise<DVCVariableSet> | void {
-        let config: BucketedUserConfig
-        if (this.config) {
-            config = this.config
-        } else {
-            throw new Error('Client not initialized')
-        }
-
         const promise = new Promise<DVCVariableSet>((resolve, reject) => {
             try {
                 this.eventQueue.flushEvents()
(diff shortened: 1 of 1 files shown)


chungchantal (approved):

sdk/js/src/Client.ts
```diff
@@ -114,7 +117,8 @@ export class DVCClient implements Client {
         if (this.config) {
             config = this.config
         } else {
-            throw new Error('Client not initialized')
+            this.userQueue.queueIdentify(user, callback)
```
ajwootto, line 120: what happens if the client is already initialized and you call "identify" multiple times. Won't that cause parallel config requests to happen? Should we queue those as well?

sdk/js/src/EventEmitter.ts
```diff
@@ -94,12 +96,17 @@ export class EventEmitter {
         })
     }
 
-    emitFeatureUpdates(oldFeatureSet: DVCFeatureSet, newFeatureSet: DVCFeatureSet): void {
-        const keys = Object.keys(oldFeatureSet).concat(Object.keys(newFeatureSet))
+    emitFeatureUpdates(
+        oldFeatureSet: DVCFeatureSet | undefined,
```
ajwootto, line 100: how come this needs to take undefined? Isn't an empty object a valid DVCFeatureSet?

jsalaber: I made the change since if the old config isn't defined yet it will be defaulted to `{}` which will make pulling the features/variables undefined, but I can make it be defined higher up so i don't have to change the signature

ajwootto (approved):

PR #121, opened 2022-04-19
fix: change initializeDVCClient to return client instead of promise

chungchantal:
- `initializeDVCClient` returns the client object instead of a promise
- update context typing so the client is not undefined in `useDVCClient`: https://medium.com/@rivoltafilippo/typing-react-context-to-avoid-an-undefined-default-value-2c7c5a7d5947

### DIFF
--- a/sdk/react/src/DVCProvider.tsx
+++ b/sdk/react/src/DVCProvider.tsx
@@ -1,8 +1,7 @@
 import { ProviderConfig } from './types'
-import React, { ReactNode, useEffect, useState } from 'react'
+import React, { ReactNode, useEffect, useMemo, useState } from 'react'
 import initializeDVCClient from './initializeDVCClient'
 import { Provider } from './context'
-import type { DVCClient } from '@devcycle/devcycle-js-sdk'
 
 type Props = {
   config: ProviderConfig
@@ -11,12 +10,8 @@ type Props = {
 
 export default function DVCProvider(props: Props): React.ReactElement {
     const { envKey, user, options } = props.config
-    const [client, setClient] = useState<DVCClient>(undefined!)
+    const client = useMemo(() => initializeDVCClient(envKey, user, options), [envKey, user, options])
     const [_, forceRerender] = useState({})
-
-    if (!client) {
-        setClient(initializeDVCClient(envKey, user, options))
-    }
 
     useEffect(() => {
         client.subscribe('variableUpdated:*', () => {
--- a/sdk/react/src/DVCProvider.tsx
+++ b/sdk/react/src/DVCProvider.tsx
@@ -11,13 +11,15 @@ type Props = {
 
 export default function DVCProvider(props: Props): React.ReactElement {
     const { envKey, user, options } = props.config
-    const [client, setClient] = useState<DVCClient | undefined>(undefined)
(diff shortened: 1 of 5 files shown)


sdk/react/src/DVCProvider.tsx
```diff
@@ -11,13 +11,15 @@ type Props = {
...
 
+    if (!client) {
+        setClient(initializeDVCClient(envKey, user, options))
+    }
+
     useEffect(() => {
         (async () => {
-            const client = await initializeDVCClient(envKey, user, options)
-            setClient(client)
             client.subscribe('variableUpdated:*', () => {
```
jsalaber, line 21: You can move this out of the async iife

jsalaber (changes requested):

sdk/react/src/DVCProvider.tsx
```diff
@@ -11,17 +11,17 @@ type Props = {
 
 export default function DVCProvider(props: Props): React.ReactElement {
     const { envKey, user, options } = props.config
-    const [client, setClient] = useState<DVCClient | undefined>(undefined)
+    const [client, setClient] = useState<DVCClient>(undefined!)
```
ajwootto, line 14: you can't default this state to undefined otherwise you're violating the type definition. It needs to be set to a valid client at all times

ajwootto (changes requested):

sdk/react/src/DVCProvider.tsx
chungchantal: Cool! Learned about `useMemo`

sdk/react/src/context.ts
```diff
@@ -2,10 +2,10 @@ import { createContext } from 'react'
 import type { DVCClient, DVCVariable } from '@devcycle/devcycle-js-sdk'
 
 interface DVCContext {
-  client?: DVCClient
+  client: DVCClient
 }
 
-const context = createContext<DVCContext>({ client: undefined })
+const context = createContext<DVCContext>({ client: undefined! })
```
ajwootto, line 8: you should avoid using `!` here as well. What you're basically doing here is telling typescript that the context will always have a client field of type "DVCClient", then basically promising the compiler here that `undefined` is actually a DVCClient.

Either you should keep the type of the context here the same or find a way to create it without passing undefined as the client

sdk/react/src/DVCProvider.tsx
```diff
@@ -11,17 +10,13 @@ type Props = {
 
 export default function DVCProvider(props: Props): React.ReactElement {
     const { envKey, user, options } = props.config
-    const [client, setClient] = useState<DVCClient | undefined>(undefined)
+    const client = useMemo(() => initializeDVCClient(envKey, user, options), [envKey, user, options])
```
ajwootto, line 13: do we actually want to create a new client when these props change? Or should we make it so only one client can be created per instance of this Provider?

chungchantal: Hmm, I’m not sure. Would those props be updated very often? It looks like the dependency array has to contain all the values referenced in the function. Or is there another way to do this so that a new client isn’t created every time those props are changed?

jsalaber (approved):

PR #122, opened 2022-04-20
feat: use assembly script lib in noedjs

ajwootto:
- use assembly script lib in nodejs sdk
- import it dynamically during initialization of the client
- update variable and feature calls to work if the client hasn't finished initializing

### DIFF
--- a/lib/shared/bucketing-assembly-script/assembly/types/platformData.ts
+++ b/lib/shared/bucketing-assembly-script/assembly/types/platformData.ts
@@ -4,7 +4,6 @@ import { getStringFromJSON } from '../helpers/jsonHelpers'
 export class PlatformData extends JSON.Obj {
     platform: string
     platformVersion: string
-    deviceModel: string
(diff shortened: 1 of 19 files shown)


lib/shared/bucketing-assembly-script/assembly/types/platformData.ts
```diff
@@ -4,7 +4,6 @@ import { getStringFromJSON } from '../helpers/jsonHelpers'
 export class PlatformData extends JSON.Obj {
     platform: string
     platformVersion: string
-    deviceModel: string
```
jonathannorris, line 7: Why did you remove it from here?

ajwootto: because deviceModel isn't platform data, it's user data. In a backend context the "device" is associated with the current user being processed, it's not a platform-level thing

jonathannorris (approved):

PR #124, opened 2022-04-25
feat: queue identify and reset calls 

jsalaber:
# Changes

- Added RequestQueue class that consolidates promises based on key
- PromiseConsolidator class that consolidates any promises made before they resolve and only resolves the latest of the same type

### DIFF
--- a/sdk/js/__tests__/RequestConsolidator.spec.js
+++ b/sdk/js/__tests__/RequestConsolidator.spec.js
@@ -1,0 +1,32 @@
+import { RequestConsolidator } from '../src/RequestConsolidator'
+
+describe('RequestConsolidator Tests', () => {
+    const requestConsolidator = new RequestConsolidator()
+
+    it('should return the latest result of the same type', async () => {
+        const promise1 = new Promise((resolve, _) => {
+            setTimeout(() => resolve(1), 200)
+        })
+        const promise2 = new Promise((resolve, _) => {
+            setTimeout(() => resolve(2), 200)
+        })
+        const resultPromise1 = requestConsolidator.queue('identify', promise1)
+        const resultPromise2 = requestConsolidator.queue('identify', promise2)
+        const results = await Promise.all([resultPromise1, resultPromise2])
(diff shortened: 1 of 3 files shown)


sdk/js/src/Client.ts
```diff
@@ -168,25 +172,26 @@ export class DVCClient implements Client {
...
-                        })
-                        const oldFeatures = oldConfig.features || {}
-                        const oldVariables = oldConfig.variables || {}
-                        this.eventEmitter.emitFeatureUpdates(oldFeatures, config.features)
-                        this.eventEmitter.emitVariableUpdates(oldVariables,
-                            config.variables, this.variableDefaultMap)
-                        resolve(config.variables || {})
-                    }).catch((e) => {
-                        reject(e)
+                this.requestConsolidator.queue('identify',
```
chungchantal, line 175: Should it be 'reset' here?

sdk/js/__tests__/RequestConsolidator.spec.js
```diff
@@ -0,0 +1,32 @@
...
+        const promise2 = new Promise((resolve, _) => {
+            setTimeout(() => resolve(2), 200)
+        })
+        const resultPromise1 = requestConsolidator.queue('identify', promise1)
+        const resultPromise2 = requestConsolidator.queue('other-type', promise2)
+        const results = await Promise.all([resultPromise1, resultPromise2])
+        expect(results[0]).toEqual(1)
+        expect(results[1]).toEqual(2)
+    })
+})
```
chungchantal, line 32: Remember to add a new line here and in RequestConsolidator.ts!

sdk/js/src/Client.ts
jsalaber: they're both an `identify` type action, and we don't want them independently being called as they save the config; we want them to be consistent.

i.e. calling `identifyUser`, then `resetUser` will always give us the bucketing for the anonymous user from the result of `resetUser` for both promises.

chungchantal (approved):

PR #134, opened 2022-05-01
prevent ssr

nikolasleblanc:


examples/react/next-js-app/pages/_app.tsx
```diff
@@ -12,4 +13,7 @@ const user = {
 function MyApp({ Component, pageProps }: AppProps) {
     return <Component {...pageProps} />
 }
-export default withDVCProvider({ envKey: ENV_KEY, user: user })(MyApp as React.FC)
+
+export default dynamic(() => Promise.resolve(withDVCProvider({ envKey: ENV_KEY, user: user })(MyApp as React.FC)), {
```
ajwootto, line 17: is this what we want our example usage of the SDK on Next to be? How come its necessary to do this?

nikolasleblanc: https://taplytics.slack.com/archives/C02DU5GTB6U/p1651432614186169 @ajwootto

PR #139, opened 2022-05-03
fix: fix examples not able to use sdk

ajwootto:


jsalaber (changes requested):
will this also be needed for the `with-provider` project?

ajwootto:
> will this also be needed for the `with-provider` project?

no, it wasnt set on that one

jsalaber (approved):

PR #143, opened 2022-05-05
fix: production configuration for react example app was missing

jsalaber:


examples/react/with-async-provider/project.json
```diff
@@ -9,6 +9,7 @@
         "{options.outputPath}"
       ],
       "configurations": {
+        "production": {},
```
ajwootto, line 12: do we actually need this to be specified? It seems to only affect things if you try to `build` the example, which you shouldn't really need to do.

ajwootto (approved):

PR #148, opened 2022-05-12
feat: save user data on identify

jsalaber:
# Changes

- new endpoint that saves data to cloud if enabled through options

### DIFF
--- a/sdk/js/src/Client.ts
+++ b/sdk/js/src/Client.ts
@@ -254,5 +254,7 @@ export class DVCClient implements Client {
 const checkIfEdgeEnabled = (enableCloudData?: boolean, config?: BucketedUserConfig) => {
     return checkIfDefined(enableCloudData) && typeof enableCloudData === 'boolean'
         ? enableCloudData
-        : config?.project?.settings?.cloudEntityData?.enabled
+        // TODO: Uncomment when settings is implemented
+        // : config?.project?.settings?.cloudEntityData?.enabled
+        : false
 }


sdk/js/src/Client.ts
```diff
@@ -40,10 +40,14 @@ export class DVCClient implements Client {
         this.requestConsolidator = new RequestConsolidator()
         this.eventEmitter = new EventEmitter()
 
+        if (options?.enableCloudData) {
```
ajwootto, line 43: this also needs to happen if the project config includes the `enableCloudEntityData` setting

sdk/js/src/Request.ts
```diff
@@ -67,6 +55,33 @@ export const post = async (
...
+        method: 'PATCH',
+        url,
+        data: body,
+        headers: baseRequestHeaders(environmentKey)
+    })
+}
+
+export const getConfigJson = async (environmentKey: string, user: DVCPopulatedUser, enableCloudData?: boolean): Promise<BucketedUserConfig> => {
+    const queryParams = `${serializeUser(user)}`
+    const url = `${BASE_URL}${HOST}${CONFIG_PATH}?envKey=${environmentKey}${queryParams && '&' + queryParams}`
```
ajwootto, line 73: you should construct an actual url here using the URL object, add pass queryParams around as a plain object

sdk/js/src/Request.ts
```diff
@@ -24,36 +24,24 @@ export const EVENT_URL = 'https://events'
...
 export const EVENTS_PATH = '/v1/events'
+export const SAVE_ENTITY_PATH = '/v1/entities'
 
-export const baseRequestHeaders = (environmentKey?: string): AxiosRequestHeaders => {
+export const baseRequestHeaders = (environmentKey?: string, enableCloudData?: boolean): AxiosRequestHeaders => {
     return {
         'Content-Type': 'application/json',
-        ...(environmentKey ? { 'Authorization': environmentKey } : {})
+        ...(environmentKey ? { 'Authorization': environmentKey } : {}),
+        ...(enableCloudData ? {'Enable-Cloud-Entity-Data': `${enableCloudData}`} : {})
```
ajwootto, line 33: what's the header for? isnt the API just expecting a query param?

jsalaber: this is off of rabia's PR since i made it off of her pr

ajwootto: ah, looks like you need to rebase then

sdk/js/src/Request.ts
jsalaber: don't see the need to pass around query params but i switched to making them use `URL` and `URLSearchParams`

sdk/js/src/Request.ts
```diff
@@ -94,6 +110,28 @@ export const publishEvents = async (
...
+    if (!user || !user.user_id) {
+        throw new Error('Missing user to save to cloud data!')
+    }
+
+    const res = await patch(
+        `${BASE_URL}${HOST}${SAVE_ENTITY_PATH}`,
+        envKey,
+        user as unknown as Record<string, unknown>
+    )
+    if (res.status !== 201) {
```
ajwootto, line 127: is 201 the only possible success code here? Isn't 200 also valid?

sdk/js/src/Client.ts
```diff
@@ -43,10 +43,15 @@ export class DVCClient implements Client {
         this.store.saveUser(this.user)
             .then(() => console.log('Successfully saved user to local storage!'))
 
-        this.onInitialized = getConfigJson(environmentKey, this.user)
+        this.onInitialized = getConfigJson(environmentKey, this.user, options?.enableCloudData)
             .then((config) => {
                 const oldConfig = this.config
                 this.config = config as BucketedUserConfig
+
+                if (options?.enableCloudData || this.config.project.settings?.cloudEntityData?.enabled) {
```
ajwootto, line 51: the logic here should be 
if enableCloudData is defined (ie. is a boolean) use its value
if not defined, use the value from the project settings

sdk/js/src/Client.ts
```diff
@@ -124,6 +130,10 @@ export class DVCClient implements Client {
                     updatedUser = new DVCPopulatedUser(user)
                 }
 
+                if (this.options?.enableCloudData) {
```
ajwootto, line 133: this should have the same check as the other one

sdk/js/src/Client.ts
```diff
@@ -132,6 +142,11 @@ export class DVCClient implements Client {
                     )
                 ).then((config) => {
                     this.config = config as BucketedUserConfig
+
+                    if (this.options?.enableCloudData || this.config?.project.settings?.cloudEntityData?.enabled) {
+                        saveEntity(this.user, this.environmentKey).then((res) => console.log(`Saved response entity! ${res}`))
```
ajwootto, line 147: how come we're saving the entity twice in the identify call here?

jsalaber: meant to replace the other one

sdk/js/src/Request.ts
jsalaber: oop copied it from the publish events endpoint. Does the same apply for the publish events endpoint? should we just look for responses under some number (maybe <400)?

sdk/js/src/Client.ts
```diff
@@ -47,6 +47,11 @@ export class DVCClient implements Client {
             .then((config) => {
                 const oldConfig = this.config
                 this.config = config as BucketedUserConfig
+
+                if (options?.enableCloudData === true || this.config.project.settings?.cloudEntityData?.enabled) {
```
ajwootto, line 51: you have to use the value of enableCloudData here if it's set. If its a boolean, use its value. If it isn't set, use the project setting

sdk/js/src/Request.ts
```diff
@@ -40,16 +41,16 @@ export const get = async (url: string): Promise<AxiosResponse> => {
...
-): Promise<BucketedUserConfig> => {
-    const queryParams = `${serializeUser(user)}${enableCloudData ? ('&enableCloudEntityData=' + enableCloudData): ''}`
-    const url = `${BASE_URL}${HOST}${CONFIG_PATH}?envKey=${environmentKey}${queryParams && '&' + queryParams}`
+export const getConfigJson = async (environmentKey: string, user: DVCPopulatedUser, enableCloudData?: boolean): Promise<BucketedUserConfig> => {
+    const queryParams = new URLSearchParams(`${serializeUser(user)}`)
+    queryParams.append('envKey', environmentKey)
+    if (enableCloudData) {
+        queryParams.append('enableCloudEntityData', 'true')
+    }
+    const url = new URL(`${BASE_URL}${HOST}${CONFIG_PATH}?${queryParams.toString()}`)
```
ajwootto, line 50: you can also set `url.searchParams` to the URLSearchParams object I think

jsalaber: it's a read only property apparently

sdk/js/src/Client.ts
```diff
@@ -132,6 +138,11 @@ export class DVCClient implements Client {
                     )
                 ).then((config) => {
                     this.config = config as BucketedUserConfig
+
+                    if (this.options?.enableCloudData === true || this.config?.project.settings?.cloudEntityData?.enabled) {
+                        saveEntity(this.user, this.environmentKey).then((res) => console.log(`Saved response entity! ${res}`))
```
ajwootto, line 143: can we fire this off in parallel to the config fetch rather than doing one and then the other? It would be faster that way

sdk/js/src/Client.ts
```diff
@@ -47,6 +47,16 @@ export class DVCClient implements Client {
             .then((config) => {
                 const oldConfig = this.config
                 this.config = config as BucketedUserConfig
+
+                const enableEdgeOption = (typeof this.options?.enableCloudData === 'boolean' 
+                    && this.options?.enableCloudData)
+                const enableEdgeFromProject = !enableEdgeOption 
+                    && this.config?.project.settings?.cloudEntityData?.enabled
+
+                if (enableEdgeOption || enableEdgeFromProject) {
```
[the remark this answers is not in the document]
ajwootto: also given that you're doing this same check in two places, probably make a function for it

ajwootto (approved):

PR #160, opened 2022-05-16
fix add react-native-get-random-values library

kaushalkapasi:
- support uuid random number generation in react-native

### DIFF
--- a/sdk/js/src/User.ts
+++ b/sdk/js/src/User.ts
@@ -1,4 +1,5 @@
 import { DVCOptions, DVCUser, JSON } from './types'
+import 'react-native-get-random-values'
 import { v4 as uuidv4 } from 'uuid'
 import * as packageJson from '../package.json'
 import UAParser from 'ua-parser-js'


sdk/js/package.json
```diff
@@ -3,5 +3,8 @@
   "version": "1.2.9",
   "description": "The DevCycle JS SDK used for feature management.",
   "author": "",
-  "license": "MIT"
+  "license": "MIT",
+  "dependencies": {
+    "react-native-get-random-values": "^1.8.0"
```
ajwootto, line 8: this should be added to the root package.json, not here

sdk/js/src/User.ts
```diff
@@ -1,4 +1,5 @@
 import { DVCOptions, DVCUser, JSON } from './types'
+import 'react-native-get-random-values'
```
ajwootto, line 2: will this work in non-react native?

kaushalkapasi: yup, this is just a polyfill

kaushalkapasi: oh wait.. we already have that library in the root package.json... hmmm

ajwootto: we may have it but it wasnt being imported in the SDK

kaushalkapasi: will fix imports and update

ajwootto (approved):

PR #163, opened 2022-05-17
Dvc 3342 edge db mvp demo

rabiaq123:
- based my PR off of Jason's https://github.com/DevCycleHQ/js-sdks/pull/158
- this code probably won't be merged into main for now, but I'm making a PR so everyone can take a look and comment on what else may / may not be needed

### DIFF
--- a/examples/js/web-elements-app/src/app/app.element.ts
+++ b/examples/js/web-elements-app/src/app/app.element.ts
@@ -3,7 +3,7 @@
 import './app.element.css'
 import { initialize } from '@devcycle/devcycle-js-sdk'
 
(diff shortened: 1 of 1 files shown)


examples/js/web-elements-app/src/app/app.element.ts
```diff
@@ -24,6 +31,8 @@ export class AppElement extends HTMLElement {
           <h1>
             <span> Hello there, </span>
             ${titleVariable.value}
+            <br><br>
+            <span> EnableEdgeDB status: ${enableStatus} </span>
```
ajwootto, line 35: is this meant to indicate that the 'cloud-data-demo' flag is enabled due to some data we got from the edgedb? Or is it saying that the edgedb itself is enabled?

If it's the former, you may want to change the wording here to say something like "Demo Flag Status"

PR #173, opened 2022-05-30
Added distinct default values for react with provider example

cristinataplytics:
Added distinct default values for react with provider example

### DIFF
--- a/examples/react/with-provider/src/DevCycleExample.tsx
+++ b/examples/react/with-provider/src/DevCycleExample.tsx
@@ -13,8 +13,7 @@ export default function DevCycleExample(): React.ReactElement {
     const variableString = useVariable( variableKeyString, 'green')
     const variableNumber = useVariable( variableKeyNumber, 600)
     const variableBoolean = useVariable( variableKeyBoolean, false)
-    //const variableJson = useVariable( variableKeyJson, { 'jsonStringKey1':'json string value 1' })
-     const variableJsonString = useVariable( variableKeyJsonString, { 'jsonStringKeyDefault':'json string value default' })
+    const variableJsonString = useVariable( variableKeyJsonString, { 'jsonStringKeyDefault':'json string value default' })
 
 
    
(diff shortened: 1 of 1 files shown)


examples/react/with-provider/src/DevCycleExample.tsx
```diff
@@ -7,13 +7,18 @@ export default function DevCycleExample(): React.ReactElement {
...
+    const variableKeyJsonString = 'variable-json-key-string'
 
     const variable = useVariable( variableKey, true)
-    const variableString = useVariable( variableKeyString, 'pink')
-    const variableNumber = useVariable( variableKeyNumber, 100)
+    const variableString = useVariable( variableKeyString, 'green')
+    const variableNumber = useVariable( variableKeyNumber, 600)
     const variableBoolean = useVariable( variableKeyBoolean, false)
-    const variableJson = useVariable( variableKeyJson, { 'jsonStringKey1':'json string value 1' })
+    //const variableJson = useVariable( variableKeyJson, { 'jsonStringKey1':'json string value 1' })
```
jsalaber, line 16: remove comments if not needed

examples/react/with-provider/src/DevCycleExample.tsx
```diff
@@ -7,13 +7,18 @@ export default function DevCycleExample(): React.ReactElement {
...
+     const variableJsonString = useVariable( variableKeyJsonString, { 'jsonStringKeyDefault':'json string value default' })
```
jsalaber, line 17: why change the name to `variableJsonString` if it's not a string?

jsalaber (changes requested):

examples/react/with-provider/src/DevCycleExample.tsx
cristinataplytics: Done, thank you!

examples/react/with-provider/src/DevCycleExample.tsx
cristinataplytics: because the Json variable value in this case is string; can be number, boolean as well; make sense for automation testing

PR #179, opened 2022-05-31
Add variation name and id to variables and features on bucketed config

emmawillis:


lib/shared/bucketing-assembly-script/assembly/types/bucketedUserConfig.ts
```diff
@@ -129,7 +134,9 @@ export class SDKVariable extends JSON.Obj {
         public type: string,
         public key: string,
         public value: JSON.Value,
-        public evalReason: string | null
+        public _variation: string,
```
ajwootto, line 137: there should be some accompanying changes somewhere that pass these new things into the constructor when building a bucketed config

lib/shared/bucketing-assembly-script/assembly/types/bucketedUserConfig.ts
```diff
@@ -106,6 +107,7 @@ export class SDKFeature extends JSON.Obj {
             getStringFromJSON(feature, 'type'),
             getStringFromJSON(feature, 'key'),
             getStringFromJSON(feature, '_variation'),
+            getStringFromJSONOptional(feature, 'variationName'),
```
ajwootto, line 110: does the variationName need to be optional? Is it not a required field in the config data?

emmawillis: ah true, it's optional in a few places in the types (like in @devcycle/types.../feature.ts) but it's required on the actual mongo schema so i think it's safe to assume it's actually required

emmawillis: oh wait never mind - we have it optional here https://github.com/DevCycleHQ/js-sdks/blob/main/lib/shared/bucketing-assembly-script/assembly/types/feature.ts#L49 so i think it needs to be optional

ajwootto: we can change that to required too. We know that variation name will always appear in the project config (due to the schema of the model), so we should change the config types to represent that

emmawillis: Makes sense, fixed!

PR #181, opened 2022-05-31
Prevent JS SDK from saving user data for anonymous users

rabiaq123:


sdk/js/__tests__/Client.spec.js
```diff
@@ -242,17 +242,30 @@ describe('DVCClient tests', () => {
...
         })
 
+        it('should not send a request to edgedb after getting the config for an anonymous user', async () => {
+            const newUser = { isAnonymous: true, country: 'CA' }
+            getConfigJson_mock.mockImplementation(() => {
+                return Promise.resolve(testConfig)
+            })
+
+            const client = new DVCClient('test_env_key', 
+                new DVCPopulatedUser({ isAnonymous: true }, { enableEdgeDB: true }),
```
ajwootto, line 260: your brackets arent quite right here

ajwootto: there's no second arg for DVCPopulatedUser

sdk/js/__tests__/Client.spec.js
```diff
@@ -242,17 +242,30 @@ describe('DVCClient tests', () => {
...
+            })
+
+            const client = new DVCClient('test_env_key', 
+                new DVCPopulatedUser({ isAnonymous: true }, { enableEdgeDB: true }), 
+                { enableEdgeDB: true }
+            )
+            await client.identifyUser(newUser)
+
+            expect(getConfigJson).toBeCalled()
+            expect(saveEntity_mock).not.toBeCalled()
```
ajwootto, line 266: a good way to make sure your test is working properly is to run it against the unchanged SDK code, and make sure it fails. Then add your fix and make sure it now passes. I have a feeling this test will always pass.

sdk/js/__tests__/Client.spec.js
```diff
@@ -242,17 +242,30 @@ describe('DVCClient tests', () => {
                 return Promise.resolve(testConfig)
             })
             saveEntity_mock.mockResolvedValue({})
-            const client = new DVCClient('test_env_key', { user_id: 'user1' }, {enableEdgeDB: true})
-            await client.onClientInitialized()
```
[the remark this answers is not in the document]
rabiaq123: whoops something might have gone wrong while i was rebasing my branch onto main... i'll check it out

jsalaber (approved):

PR #193, opened 2022-06-07
DVC 3633 - WASM memory debugging

JamieSinn:


lib/shared/bucketing-assembly-script/hotload.js
```diff
@@ -0,0 +1,4 @@
+import * as debug from 'build/bucketing-lib.debug'
+import * as release from 'build/bucketing-lib.release'
+
+export const allExports = !!Number(process.env.DEVCYCLE_USE_DEBUG_WASM) ? debug : release
```
ajwootto, line 4: doesn't this change the interface of this package?

you would now need to import it like this:
`import { allExports } from '@devcycle/bucketing-assembly-script`

did the "entrypoints" approach not work?

JamieSinn: entrypoints isn't supported it seemed for the way we wanted to do it - it was saying the option is unsupported.

is there a way to just raw export from the package? or do we now have to pass through an export for all the sub-properties/functions?

ajwootto: well its an esmodule so the exports have to be named upfront, and in our case we want it to export each method as a named property not have it all be inside something like "allExports"

PR #200, opened 2022-06-14
DVC-2234: create new hook that awaits for DVC to initialize

chungchantal:
- using useState to rerender the app when `isDVCReady` is true
- sets to true when catching error from `onClientInitialized()` to unblock the app from loading

### DIFF
--- a/sdk/react/src/useDVCInitialized.ts
+++ b/sdk/react/src/useDVCInitialized.ts
@@ -1,0 +1,24 @@
+import { useContext, useState } from 'react'
+import context from './context'
+
+export const useDVCInitialized = (): boolean => {
+    const [isDVCReady, setIsDVCReady] = useState(false)
+    if (isDVCReady) return isDVCReady
+
+    const dvcContext = useContext(context)
+
+    if (dvcContext === undefined) throw new Error('useDVCInitialized must be used within DVCProvider')
+
+    dvcContext.client.onClientInitialized()
+    .then(() => {
+        setIsDVCReady(true)
+    }).catch(() => {
+        // set to true to unblock app load
+        console.log('Error initializing DevCycle.')
+        setIsDVCReady(true)
+    })
+
+    return isDVCReady
+}
+
+export default useDVCInitialized
--- a/sdk/react/src/useDVCInitialized.ts
+++ b/sdk/react/src/useDVCInitialized.ts
@@ -3,11 +3,11 @@ import context from './context'
 
 export const useDVCInitialized = (): boolean => {
     const [isDVCReady, setIsDVCReady] = useState(false)
-    if (isDVCReady) return isDVCReady
-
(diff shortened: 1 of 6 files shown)


sdk/react/src/useDVCInitialized.ts
```diff
@@ -0,0 +1,24 @@
+import { useContext, useState } from 'react'
+import context from './context'
+
+export const useDVCInitialized = (): boolean => {
+    const [isDVCReady, setIsDVCReady] = useState(false)
+    if (isDVCReady) return isDVCReady
+
+    const dvcContext = useContext(context)
```
squarepegg, line 8: Should this line and the undefined check be moved to the top of the function so React will render the same way on every call?

sdk/react/src/useDVCInitialized.ts
```diff
@@ -0,0 +1,24 @@
...
+
+    if (dvcContext === undefined) throw new Error('useDVCInitialized must be used within DVCProvider')
+
+    dvcContext.client.onClientInitialized()
```
squarepegg, line 12: Where are the env key and start up options passed now?

sdk/react/src/useDVCInitialized.ts
chungchantal: Yup, I'll update :) the linter also caught this!

sdk/react/src/useDVCInitialized.ts
```diff
@@ -0,0 +1,24 @@
+import { useContext, useState } from 'react'
+import context from './context'
+
+export const useDVCInitialized = (): boolean => {
```
squarepegg, line 4: Does this need to be awaitable? So that 2 modes are available to the user?

if awaited then the rendering is blocked

If not awaited then the app loads but the client is usable with default variables if not initialized

chungchantal: That's what we used to have with `asyncWithDVCProvider` but I think we want to be able to use this hook like so:
```
export const MyApp = () => {
  const dvcReady = useDVCInitialized()

  if (!dvcReady) return <LoadingState/>

  return <TheRestofYourApp/>
}

chungchantal: You could just use `withDVCProvider` wrapper on its own without this new `useDVCInitialized` hook!

squarepegg: kk

sdk/react/src/useDVCInitialized.ts
squarepegg: kk

squarepegg (commented):
Last comment? :D

Does this need a test / example app usage?

squarepegg (approved):

ajwootto (approved):

PR #202, opened 2022-06-15
DVC Logger Types & Log level customization

kaushalkapasi:
- updates JS and NodeJS sdks to use DVC Logger functions defined within each SDK
- updates `@devcycle/types` to have reference to DVC Logger types


examples/js/web-elements-app/src/app/app.element.ts
```diff
@@ -452,7 +453,7 @@ const user = {
     isAnonymous: false
 }
 
-const options = { enableEdgeDB: true }
+const options = { enableEdgeDB: true, logLevel: 'error' as DVCDefaultLogLevel }
```
ajwootto, line 456: casting the logLevel value like this can't be a requirement for Typescript users. Using the string directly should be possible

lib/shared/logger/project.json
```diff
@@ -0,0 +1,37 @@
...
+      "outputs": [
+        "{options.outputPath}"
+      ],
+      "options": {
+        "outputPath": "dist/lib/shared/logger",
+        "main": "lib/shared/logger/src/index.ts",
+        "tsConfig": "lib/shared/logger/tsconfig.lib.json"
+      }
+    },
+    "build:json": {
```
ajwootto, line 28: this isnt a valid target for this lib

examples/js/web-elements-app/src/app/app.element.ts
```diff
@@ -452,7 +452,5 @@ const user = {
     isAnonymous: false
 }
 
-const options = { enableEdgeDB: true }
-
-const client = initialize(clientKey, user, options)
+const client = initialize(clientKey, user, { enableEdgeDB: true, logLevel: 'error' })
```
ajwootto, line 455: edgedb shouldnt be enabled in this example

package.json
```diff
@@ -133,6 +133,7 @@
     "react-test-renderer": "18.0.0",
     "service-worker-mock": "^2.0.5",
     "ts-jest": "27.1.4",
+    "ts-node": "9.1.1",
```
ajwootto, line 136: whats this being used for?

kaushalkapasi: ah, need to remove that.. missed the cleanup from the unit test debugging

sdk/js/src/Client.ts
```diff
@@ -39,24 +42,25 @@ export class DVCClient implements Client {
         this.eventQueue = new EventQueue(environmentKey, this, options?.flushEventsMS)
         this.requestConsolidator = new RequestConsolidator()
         this.eventEmitter = new EventEmitter()
+        this.logger = options?.logger || dvcDefaultLogger({ level: options?.logLevel })
```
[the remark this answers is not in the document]
ajwootto: actually I guess it could be if someone passes null in explicitly while using plain Javascript. Should probably just validate all the constructor properties further up though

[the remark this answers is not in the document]
kaushalkapasi: should we throw an error if options are undefined / null or just log that options are undefined and then create default options?

sdk/js/src/Store.ts
```diff
@@ -37,7 +36,7 @@ export class Store {
                 resolve(modify(storeKey))
             })
         } catch (e) {
-            console.log(e)
+            console.error(e)
```
jsalaber, line 39: should this be `logger.error`?

jsalaber (changes requested):

sdk/js/src/Store.ts
kaushalkapasi: updated

jsalaber (approved):

sdk/js/src/index.ts
```diff
@@ -27,12 +27,16 @@ export const initialize = (environmentKey: string, user: DVCUser, options: DVCOp
         throw new Error('Missing environment key! Call initialize with a valid environment key')
     }
 
+    if (!options || options === null) {
```
ajwootto, line 30: should we check the user too

sdk/js/src/Client.ts
```diff
@@ -31,32 +34,33 @@ export class DVCClient implements Client {
...
     constructor(environmentKey: string, user: DVCPopulatedUser, options: DVCOptions = {}) {
-        this.store = new Store(window.localStorage)
         this.user = user
         this.options = options
         this.environmentKey = environmentKey
         this.variableDefaultMap = {}
         this.eventQueue = new EventQueue(environmentKey, this, options?.flushEventsMS)
         this.requestConsolidator = new RequestConsolidator()
         this.eventEmitter = new EventEmitter()
+        this.logger = options?.logger || dvcDefaultLogger({ level: options?.logLevel })
```
ajwootto, line 44: options has to exist so get rid of the optional accessor

sdk/js/src/index.ts
kaushalkapasi: yeah, will add another check and throw and error if the user doesn't meet the minimum required by the `DVCUser` type

ajwootto (approved):

PR #205, opened 2022-06-22
Update nodejs example app README's with nx instructions

jonathannorris:


JamieSinn (approved):

kaushalkapasi (approved):

ajwootto (commented):
These commands would only work if you have nx installed globally. You should prefix them with "yarn" or "npx" to use the repo-installed nx executable, or add a script to package.json if there isn't one already

ajwootto (approved):

PR #207, opened 2022-06-22
fix entry file path for react native android

chungchantal:
- the command `yarn nx run-android example-react-native-app` wouldn't work because the entry file path was wrong

### DIFF
--- a/examples/react/react-native-app/android/app/build.gradle
+++ b/examples/react/react-native-app/android/app/build.gradle
@@ -79,7 +79,7 @@ import com.android.build.OutputFile
  */
 
 project.ext.react = [
-    entryFile: "examples/react/react-native-app/src/main.tsx",
+    entryFile: "src/main.tsx",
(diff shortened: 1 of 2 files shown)


examples/react/react-native-app/android/app/build.gradle
```diff
@@ -79,7 +79,7 @@ import com.android.build.OutputFile
  */
 
 project.ext.react = [
-    entryFile: "examples/react/react-native-app/src/main.tsx",
```
ajwootto, line 82: are you running the nx commands from the root or from the example app's directory? that could explain why it wouldn't be working

chungchantal: ah I was running it from the example apps directory, not root. so nvm will close this!
(I had tested by regenerating a new RN app thru nx and it has the entry file set to `src/main.tsx`)

PR #208, opened 2022-06-23
Transform user string date format into Date obj

emmawillis:


lib/shared/types/src/types/apis/sdk/clientSDKAPI.ts
```diff
@@ -111,15 +120,15 @@ export class DVCAPIUser {
      */
     @IsDate()
     @IsOptional()
-    @Transform(({ value }) => value ? new Date(Number(value)) : undefined)
+    @Transform(dateTransform)
```
ajwootto, line 123: you should just be able to remove the `Number()` call here and pass the value directly into the Date constructor.

ajwootto (approved):

PR #212, opened 2022-06-28
[DVC-3403] variableDefaulted logging

rabiaq123:
- send `variableDefaulted` event only when variable is NOT in bucketed config

### DIFF
--- a/sdk/nodejs/__tests__/client.spec.ts
+++ b/sdk/nodejs/__tests__/client.spec.ts
@@ -18,3 +18,34 @@ describe('DVCClient', () => {
         })
     })
 })
+
+describe('variable', () => {
+    const user = {
+        user_id: 'node_sdk_test',
+        country: 'CA'
+    }
+    const expectedUser = expect.objectContaining({
+        user_id: 'node_sdk_test',
+        country: 'CA'
+    })
+    
+    describe('variableDefaulted event', () => {
+        test('does not get sent if variable is in bucketed config',async () => {
+            const client = new DVCClient('token')
+
+            await client.onClientInitialized()
+            // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+            // @ts-ignore
+            client.eventQueue.queueAggregateEvent = jest.fn()
+            client.variable(user, 'test-key', false)
+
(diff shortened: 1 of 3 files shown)


jsalaber (approved):

sdk/nodejs/__tests__/client.spec.ts
```diff
@@ -18,3 +18,34 @@ describe('DVCClient', () => {
...
+        user_id: 'node_sdk_test',
+        country: 'CA'
+    }
+    const expectedUser = expect.objectContaining({
+        user_id: 'node_sdk_test',
+        country: 'CA'
+    })
+    
+    describe('variableDefaulted event', () => {
+        test('does not get sent if variable is in bucketed config',async () => {
```
ajwootto, line 33: you should add another test case for "it does get sent when the variable is not in the bucketed config"

rabiaq123: oh i did initially, but that test also passed w the previous code, so i removed it.
should i make it so that it fails w the prev code? or is it ok if it passes there too?

ajwootto: it would have passed before as well because the old behaviour would still have sent `variableDefaulted` if the variable wasn't in the config. The bug you fixed was just for the case where it _was_ in the config. However, there isn't an existing test to cover the fact that variableDefaulted is being sent, and since you changed the code that sends it it would be a good idea to add one to make sure that still works.

sdk/nodejs/__tests__/client.spec.ts
```diff
@@ -18,3 +18,52 @@ describe('DVCClient', () => {
...
+            await client.onClientInitialized()
+            // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+            // @ts-ignore
+            client.eventQueue.queueAggregateEvent = jest.fn()
+        })
+        beforeEach(() => {
+            jest.clearAllMocks()
+        })
+    
+        test('does not get sent if variable is in bucketed config',async () => {
```
ajwootto, line 46: generally we use `it` instead of `test`

ajwootto (approved):

PR #213, opened 2022-06-29
fix: use pattern to block dist folders not under node modules

chantalchung:
- the regex pattern suggested in this issue https://github.com/nrwl/nx/issues/10644 allows us to build the react native example app when the root level dist folder has content


ajwootto (approved):

ajwootto:
if this works and you've tested it, wanna comment on that issue letting them know that it seems to work?

chungchantal:
@ajwootto Yup I'll leave a comment on that issue!

PR #215, opened 2022-06-30
Enable EdgeDB for NodeJS SDK

kaushalkapasi:
- adds `enableEdgeDB` to DVCOptions
- checks is `enableEdgeDB` option is set to true && `enableCloudBucketing` is set to true, to pass through Enable EdgeDB to Bucketing API
  - creates INFO level log, if `enableEdgeDB` is set to true without Cloud Bucketing

### DIFF
--- a/sdk/nodejs/src/client.ts
+++ b/sdk/nodejs/src/client.ts
@@ -40,6 +40,10 @@ export class DVCClient {
         this.options = options
         this.logger = options?.logger || dvcDefaultLogger({ level: options?.logLevel })
 
+        if (options?.enableEdgeDB) {
+            this.logger.info('EdgeDB can only be enabled for the DVC Cloud Client.')
+        }
+
         const initializePromise = importBucketingLib()
             .then(() => {
                 this.configHelper = new EnvironmentConfigManager(this.logger, environmentKey, options || {})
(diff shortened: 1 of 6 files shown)


sdk/nodejs/src/client.ts
```diff
@@ -40,6 +40,10 @@ export class DVCClient {
         this.options = options
         this.logger = options?.logger || dvcDefaultLogger({ level: options?.logLevel })
 
+        if (options?.enableEdgeDB) {
+            this.logger.info('EdgeDB can only be enabled for the DVC Cloud Client.')
```
squarepegg, line 44: Should this be a warning?

sdk/nodejs/src/cloudClient.ts
```diff
@@ -32,7 +34,7 @@ export class DVCCloudClient {
         checkParamDefined('key', key)
         checkParamDefined('defaultValue', defaultValue)
 
-        return getVariable(requestUser, this.environmentKey, key)
+        return getVariable(requestUser, this.environmentKey, key, this.options.enableEdgeDB)
```
squarepegg, line 37: Could options be null / undefined?

kaushalkapasi: `options` won't be... but enableEdgeDB might be undefined.. let me write a test for that usecase

sdk/nodejs/src/client.ts
kaushalkapasi: I don't believe so, this will not prevent the SDK from functioning correctly... it will simply not leverage the Edge DB functionality, so I made it an info level log

squarepegg (approved):

PR #221, opened 2022-07-15
Flush events if max event queue size has been reached

rabiaq123:


sdk/nodejs/src/eventQueue.ts
```diff
@@ -131,6 +135,12 @@ export class EventQueue {
 
     private addEventToQueue(user: DVCPopulatedUser, event: DVCRequestEvent) {
         let userEvents = this.userEventQueue[user.user_id]
+
+        this.eventQueueLength++
```
ajwootto, line 139: you'll have to set this back to zero at some point when the events are flushed

jsalaber (dismissed):
maybe add a unit test?

PR #222, opened 2022-07-15
feat: added a max event queue size for storing events

jsalaber:
# Changes

- added function to find equivalent event queue size for aggregate events and user events
- drops events and aggregate events from queueing if max event queue size limit is reached
- add type for future use of configurable event queue size from bucketed config

### DIFF
--- a/sdk/nodejs/__tests__/eventQueue.spec.ts
+++ b/sdk/nodejs/__tests__/eventQueue.spec.ts
@@ -18,7 +18,12 @@ describe('EventQueue Unit Tests', () => {
         environment: { _id: '', key: '' },
         features: {},
         knownVariableKeys: [],
-        project: { _id: '', key: '', a0_organization: 'org_' } as PublicProject,
+        project: { 
+            _id: '', 
+            key: '', 
+            a0_organization: 'org_', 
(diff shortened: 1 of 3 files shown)


sdk/nodejs/__tests__/eventQueue.spec.ts
```diff
@@ -381,4 +386,46 @@ describe('EventQueue Unit Tests', () => {
...
+        for (let i = 0; i < 500; i++) {
+            eventQueue.queueAggregateEvent(new DVCPopulatedUser({ user_id: `user${i}` }), aggEvent, config)
+        }
+
+        for (let i = 0; i < 501; i++) {
+            eventQueue.queueEvent(user, { type: 'test_event' }, config)
+        }
+
+        expect(logger.warn).toBeCalledTimes(1)
+        expect(logger.warn).toBeCalledWith(expect.stringContaining('Max event queue size reached, dropping event'))
```
ajwootto, line 407: any way to assert that the events weren't actually added here?

you can access private properties of the class if you override the type warning

jsalaber: done

ajwootto (approved):

PR #224, opened 2022-07-20
Change AssemblyScript complier optimization level 

jonathannorris:
change the compiler to optimize for execution speed not bundle size

https://taplytics.slack.com/archives/CDS47F4V6/p1657860682680499

image (14)

### DIFF
--- a/lib/shared/bucketing-assembly-script/project.json
+++ b/lib/shared/bucketing-assembly-script/project.json
@@ -50,10 +50,10 @@
       "options": {
         "commands": [
           {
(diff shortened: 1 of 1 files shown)


JamieSinn (approved):

MaxGraey:
'-O3' is unnecessary for debug builds. It also unnecessary for AS version after 0.19.x due to `--optimize` now equivalent to `-O3` instead `-O3s`. Also if you're using вуафгде `asconfig` (And from the looks of it, it really is) you don't need extra `-O3` on cli's build scripts.

PR #225, opened 2022-07-20
NodeJS SDK: make enableCloudBucketing param optional & add default value

kaushalkapasi:
- make enableCloudBucketing param optional
- default to false for NodeJS SDK to ensure local bucketing is the default functionality when the option isn't present

### DIFF
--- a/sdk/nodejs/src/index.ts
+++ b/sdk/nodejs/src/index.ts
@@ -6,7 +6,7 @@ export { DVCClient, DVCCloudClient }
 export * from './types'
 
 type DVCOptionsCloudEnabled = DVCOptions & { enableCloudBucketing: true }
-type DVCOptionsLocalEnabled = DVCOptions & { enableCloudBucketing: false }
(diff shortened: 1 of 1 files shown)


sdk/nodejs/src/index.ts
```diff
@@ -6,7 +6,7 @@ export { DVCClient, DVCCloudClient }
 export * from './types'
 
 type DVCOptionsCloudEnabled = DVCOptions & { enableCloudBucketing: true }
-type DVCOptionsLocalEnabled = DVCOptions & { enableCloudBucketing: false }
+type DVCOptionsLocalEnabled = DVCOptions & { enableCloudBucketing?: false }
```
ajwootto, line 9: did you add the default value somewhere?

kaushalkapasi: this will set the default when the value is not defined, verified against the example app

ajwootto (approved):

PR #226, opened 2022-07-20
Add Name, Description and Settings to SDK Feature

kaushalkapasi:
- adds settings to SDK Feature on the User Config if it exists on the Project Config
- does not have the settings key if the `feature.settings.optInEnabled` is false

### DIFF
--- a/lib/shared/bucketing/src/bucketing.ts
+++ b/lib/shared/bucketing/src/bucketing.ts
@@ -163,7 +163,7 @@ export const generateBucketedConfig = (
     const segmentedFeatures = getSegmentedFeatureDataFromConfig({ config, user })
 
     segmentedFeatures.forEach(({ feature, target }) => {
-        const { _id, key, type, variations } = feature
+        const { _id, key, type, variations, name, description, settings } = feature
         const { rolloutHash, bucketingHash } = generateBoundedHashes(user.user_id, target._id)
         if (target.rollout && !doesUserPassRollout({ boundedHash: rolloutHash, rollout: target.rollout })) {
             return
@@ -181,7 +181,10 @@ export const generateBucketedConfig = (
             type,
             _variation: variation_id,
             variationName: variation.name,
-            variationKey: variation.key
+            variationKey: variation.key,
+            name,
+            description,
+            settings
         }
         featureVariationMap[_id] = variation_id
         variation.variables.forEach(({ _var, value }) => {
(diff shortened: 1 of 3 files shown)


lib/shared/bucketing-assembly-script/assembly/types/bucketedUserConfig.ts
```diff
@@ -125,6 +132,15 @@ export class SDKFeature extends JSON.Obj {
         if (this.evalReason) {
             json.set('evalReason', this.evalReason)
         }
+        if (this.name) {
+            json.set('name', this.name)
+        }
+        if (this.description) {
+            json.set('evalReason', this.description)
```
ajwootto, line 139: evalReason?

kaushalkapasi: oops.. good catch!

lib/shared/bucketing-assembly-script/assembly/bucketing/bucketing.ts
```diff
@@ -222,7 +222,10 @@ export function _generateBucketedConfig(
             variation_id,
             variation.name,
             variation.key,
-            null
+            null,
+            feature.name,
+            feature.description,
```
jonathannorris, line 227: Why would we want these server-side? this is going to be expensive in bandwidth to send down to configs.

kaushalkapasi: ah.. right, Bucketing-Assembly-Script is only used Server Side so we wouldn't need this since the user will never be able to enrol into a Beta / Opt-In feature program on the Server Side, only on client side

jonathannorris: Even client-side we probably shouldn't be including that in all configs

jonathannorris: also exposing internal data like that in our config responses client-side probably is a bad idea.

kaushalkapasi: would it make sense to do it the same way as the config generation, conditional checking based on if the `feature.settings.optInEnabled` is true, then add these keys? 🤔

jonathannorris (approved):

PR #233, opened 2022-07-21
add getOptInSettings to client

rodeora:


sdk/js/src/types.ts
```diff
@@ -178,6 +189,11 @@ export interface DVCClient {
      */
     allVariables(): DVCVariableSet
 
+    /**
+     * Retrieve configuration settings related to Opt-in
+     */
+    getOptInSettings(): OptInSettings
```
kaushalkapasi, line 195: we should be able to change this to:
```suggestion
    getOptInSettings(): Project["settings"]["optIn"]
```
instead of creating a new type definition

PR #234, opened 2022-07-22
Add feature settings, user optIns, and optIn filter to types

emmawillis:


lib/shared/bucketing-assembly-script/assembly/types/configBody.ts
```diff
@@ -12,24 +12,22 @@ export class PublicProject extends JSON.Value {
...
-    settings: JSON.Obj | null
+    settings: JSON.Obj
 
     constructor(project: JSON.Obj) {
         super()
         this._id = getStringFromJSON(project, '_id')
         this.key = getStringFromJSON(project, 'key')
         this.a0_organization = getStringFromJSON(project, 'a0_organization')
-        this.settings = getJSONObjFromJSONOptional(project, 'settings')
+        this.settings = getJSONObjFromJSON(project, 'settings')
```
ajwootto, line 22: does every config in the CDN have a settings field now?

lib/shared/bucketing-assembly-script/assembly/types/dvcUser.ts
```diff
@@ -26,6 +26,7 @@ export class DVCUser extends JSON.Obj implements DVCUserInterface {
     appVersion: string | null
     appBuild: f64
     customData: JSON.Obj | null
+    optIns: JSON.Obj | null
```
ajwootto, line 29: this isn't something that can be passed in is it? If not it shouldn't be on DVCUser

emmawillis: yeah but that would require an sdk update for all the sdks that use this. the idea was that the optIn data is stored in edgeDB as user.optIns, so all the places where we combine edgeDB data with user data would automatically handle it by adding it to the DVCUser object that gets sent for bucketing, without needing an update

lib/shared/bucketing-assembly-script/assembly/types/configBody.ts
emmawillis: https://github.com/DevCycleHQ/js-sdks/blob/main/lib/shared/types/src/types/config/models/project.ts#L12 we made it required here already and treat it as required in other places, so I think so

lib/shared/types/src/types/apis/sdk/clientSDKAPI.ts
```diff
@@ -116,6 +116,13 @@ export class DVCAPIUser {
     @Transform(({ value }) => isString(value) ? JSON.parse(value) : value)
         privateCustomData?: DVCJSON
 
+    /**
+     * JSON data recording user opt-in features.
+     */
+     @IsOptional()
+     @Transform(({ value }) => isString(value) ? JSON.parse(value) : value)
+         optIns?: Record<string, boolean>
```
ajwootto, line 124: this can't be passed to the API can it?

kaushalkapasi (approved):

PR #235, opened 2022-07-22
Handle optIn filter in bucketing and ignore in bucketing-as

emmawillis:


lib/shared/bucketing/src/segmentation.ts
```diff
@@ -15,11 +15,18 @@ import UAParser from 'ua-parser-js'
...
+export const evaluateOperator = (
+    { operator, data, featureId, isOptInEnabled }:
+    {operator: TopLevelOperator, data: any, featureId: string, isOptInEnabled: boolean}
+): boolean => {
     if (!operator.filters.length) return false
 
     const doesUserPassFilter = (filter: AudienceFilterOrOperator) => {
         if (filter.type === 'all') return true
+        if (filter.type === 'optIn') {
+            const optIns = data.optIns as Record<string, boolean>
```
ajwootto, line 27: how come the casting is needed here?

ajwootto (approved):

lib/shared/bucketing/src/segmentation.ts
emmawillis: not needed anymore, fixed!

PR #236, opened 2022-07-22
Add worker proxy option to SDKs

jonathannorris:
Add a new `DVCOption` `workerProxyURL` that allows the SDKs to use a local proxy of our workers to serve the requests from the SDKs.

Also split up the nodeJS tests from just `js`/`ts` example apps into `nodejs-cloud` and `nodejs-local` examples that each contain the `js`/`ts` example apps.


sdk/js/src/types.ts
```diff
@@ -47,6 +47,7 @@ export interface DVCOptions {
     enableEdgeDB?: boolean
     logger?: DVCLogger
     logLevel?: DVCDefaultLogLevel
+    workerProxyURL?: string
```
ajwootto, line 50: should this option be named something more user-facing? The fact that we're using "workers" is sort of an implementation detail. Maybe `cloudProxyURL` or `apiProxyURL`?

jonathannorris: yea `apiProxyURL` could make sense.

jsalaber (approved):

PR #237, opened 2022-07-25
fix bucketing assembly script to log errors when a new filter type is passed to segmentation

kaushalkapasi:
- remove thrown errors in Segmentation code
- instead the code will log warnings in Segmentation code when an unexpected filter type is received

### DIFF
--- a/lib/shared/bucketing-assembly-script/assembly/bucketing/segmentation.ts
+++ b/lib/shared/bucketing-assembly-script/assembly/bucketing/segmentation.ts
@@ -1,13 +1,13 @@
 import { RegExp } from 'assemblyscript-regex/assembly'
-import {  findString, includes, replace } from '../helpers/lodashHelpers'
+import { findString, includes, replace } from '../helpers/lodashHelpers'
 import { OptionsType, versionCompare } from './versionCompare'
 import {
-    TopLevelOperator, 
-    AudienceFilterOrOperator, 
-    DVCPopulatedUser, 
-    validSubTypes, 
+    TopLevelOperator,
+    AudienceFilterOrOperator,
+    DVCPopulatedUser,
+    validSubTypes,
     CustomDataFilter,
-    UserFilter, 
+    UserFilter,
 } from '../types'
 import { JSON } from 'assemblyscript-json/assembly'
 import { getF64FromJSONValue } from '../helpers/jsonHelpers'
@@ -20,7 +20,7 @@ import { getF64FromJSONValue } from '../helpers/jsonHelpers'
  * @param user - The incoming user, device, and user agent data
  */
 export function _evaluateOperator(
-    operator: TopLevelOperator, 
+    operator: TopLevelOperator,
     user: DVCPopulatedUser
 ): bool {
     if (!operator.filters.length) return false
@@ -47,20 +47,28 @@ export function _evaluateOperator(
 }
 
 function doesUserPassFilter(
-    filter: AudienceFilterOrOperator, 
+    filter: AudienceFilterOrOperator,
     user: DVCPopulatedUser
 ): bool {
     if (filter.type === 'all') return true
     if (filter.type === 'optIn') return false
     if (!(filter instanceof UserFilter)) {
-        throw new Error('Invalid filter data')
+        console.log(`
+            [DevCycle] Error: Invalid filter data ${filter}.
+            To leverage this new filter definition, please update to the latest version of the DevCycle SDK in use
+        `)
+        return false
     }
 
     const userFilter = filter as UserFilter
 
     const subType = userFilter.subType
     if (!validSubTypes.includes(subType)) {
-        throw new Error(`Invalid filter subType: ${subType}`)
+        console.log(`
+            [DevCycle] Error: Invalid filter subType: ${subType}.
+            To leverage this new filter definition, please update to the latest version of the DevCycle SDK in use
+        `)
+        return false
     }
 
     return filterFunctionsBySubtype(subType, user, userFilter)
@@ -153,7 +161,7 @@ export function checkVersionFilter(
 
         const mappedFilterVersions: string[] = []
         // Replace Array.map(), because you can't access captured data in a closure
-        for (let i=0; i < filterVersions.length; i++) {
+        for (let i = 0; i < filterVersions.length; i++) {
             mappedFilterVersions.push(replace(replace(filterVersions[i], regex1, ''), regex2, ''))
         }
         parsedFilterVersions = mappedFilterVersions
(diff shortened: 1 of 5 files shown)


lib/shared/bucketing-assembly-script/assembly/bucketing/segmentation.ts
```diff
@@ -83,7 +91,11 @@ function filterFunctionsBySubtype(subType: string, user: DVCPopulatedUser, filte
         return _checkStringsFilter(user.platform, filter)
     } else if (subType === 'customData') {
         if (!(filter instanceof CustomDataFilter)) {
-            throw new Error('Invalid filter data')
```
ajwootto, line 86: I think this would legitimately be an error, it shouldn't be able to reach this point without the filter being a CustomDataFilter. This was mostly here for the benefit of the type system I think

lib/shared/bucketing-assembly-script/assembly/bucketing/segmentation.ts
```diff
@@ -47,20 +47,28 @@ export function _evaluateOperator(
...
 function doesUserPassFilter(
-    filter: AudienceFilterOrOperator, 
+    filter: AudienceFilterOrOperator,
     user: DVCPopulatedUser
 ): bool {
     if (filter.type === 'all') return true
     if (filter.type === 'optIn') return false
     if (!(filter instanceof UserFilter)) {
-        throw new Error('Invalid filter data')
+        console.log(`
```
ajwootto, line 56: can you write tests for this

lib/shared/bucketing-assembly-script/assembly/bucketing/segmentation.ts
kaushalkapasi: will revert this one

lib/shared/bucketing-assembly-script/assembly/bucketing/segmentation.ts
```diff
@@ -47,20 +47,28 @@ export function _evaluateOperator(
...
+            [DevCycle] Error: Invalid filter data ${filter}.
+            To leverage this new filter definition, please update to the latest version of the DevCycle SDK in use
```
ajwootto, line 58: not sure what `in use` means at the end here. Maybe just drop that?

lib/shared/bucketing-assembly-script/assembly/helpers/jsonHelpers.ts
```diff
@@ -46,6 +46,27 @@ export function isValidString(jsonObj: JSON.Obj, key: string, validStrings: stri
     return value
 }
 
+export function isUsableString(jsonObj: JSON.Obj, key: string, validStrings: string[]): string {
+    const str = getStringFromJSON(jsonObj, key)
+    if (!validStrings.includes(str)) {
+        console.log(`[DevCycle] Warning: String value: ${str}, for key: ${key} does not match a valid string.`)
```
ajwootto, line 52: should this be using `console.warn`?

lib/shared/bucketing-assembly-script/assembly/helpers/jsonHelpers.ts
```diff
@@ -46,6 +46,27 @@ export function isValidString(jsonObj: JSON.Obj, key: string, validStrings: stri
     return value
 }
 
+export function isUsableString(jsonObj: JSON.Obj, key: string, validStrings: string[]): string {
```
ajwootto, line 49: the names of these methods are a bit weird. Unclear what the difference between `valid` and `usable` is, but I can't really think of a better distinction

kaushalkapasi: the other option I can think of is `isPassableString`, such that it won't error but just passthrough and return it?

ajwootto: what about a "shouldThrow" parameter for the existing "valid" methods?

lib/shared/bucketing-assembly-script/assembly/bucketing/segmentation.ts
kaushalkapasi: since Bucketing ASM can be used across different SDKs, but yeah it doesn't make much sense reading it back

lib/shared/bucketing-assembly-script/assembly/helpers/jsonHelpers.ts
kaushalkapasi: I wasn't able to find any examples that actually use the other log levels for WebAssembly, that's why I left it as a `console.log`, but let me check again

lib/shared/bucketing-assembly-script/assembly/helpers/jsonHelpers.ts
kaushalkapasi: yeah, that sounds better... let me update accordingly

lib/shared/bucketing-assembly-script/assembly/helpers/jsonHelpers.ts
kaushalkapasi: updated to `console.warn`, it'll now look lik:
image

lib/shared/bucketing-assembly-script/assembly/helpers/jsonHelpers.ts
```diff
@@ -38,23 +38,43 @@ export function getStringFromJSONOptional(jsonObj: JSON.Obj, key: string): strin
...
+): string {
     const value = getStringFromJSON(jsonObj, key)
     if (!validStrings.includes(value)) {
-        throw new Error(`Invalid string value: ${value}, for key: ${key}, must be one of: ${validStrings.join(', ')}`)
+        if (shouldThrow) {
+            throw new Error(
+                `Invalid string value: ${value}, for key: ${key}, must be one of: ${validStrings.join(', ')}`
+            )
+        } else {
+            console.warn(`[DevCycle] Warning: String value: ${value}, for key: ${key} does not match a valid string.`)
```
JamieSinn, line 54: We should use log here. Warn will sigkill.

JamieSinn: er - more accurately - warn will do different things per impl.

ajwootto: why would it do that?

JamieSinn: depends on the WASI level of implementation. All these calls are bindings to WASI.

ajwootto: are the implementations of warn any more divergent than `log`?

JamieSinn: My understanding is that warn is generally lumped in with error for most implementations that don't have full WASI bindings (ie PHP) - but for languages with full WASI support (C#, Go, Node) it's fine and can be handled separately

kaushalkapasi: going to revert back to just `console.log` then

JamieSinn (approved):

ajwootto (approved):

PR #238, opened 2022-08-03
DVC-2495: add value, type, and _variable to event metadata

chungchantal:
- mongo object Id is only available for the `variableEvaluated` events

### DIFF
--- a/sdk/js/src/Client.ts
+++ b/sdk/js/src/Client.ts
@@ -123,6 +123,11 @@ export class DVCClient implements Client {
                     ? EventTypes.variableDefaulted
                     : EventTypes.variableEvaluated,
                 target: variable.key,
+                metaData: {
+                    value: variable.value,
+                    type: typeof(variable.defaultValue),
+                    _variable: this.config?.variables?.[key]._id
+                }
             })
         } catch (e) {
             this.eventEmitter.emitError(e)
--- a/sdk/js/src/Client.ts
+++ b/sdk/js/src/Client.ts
@@ -15,7 +15,7 @@ import { DVCPopulatedUser } from './User'
 import { EventQueue, EventTypes } from './EventQueue'
 import { checkParamDefined } from './utils'
 import { EventEmitter } from './EventEmitter'
-import { BucketedUserConfig } from '@devcycle/types'
+import { BucketedUserConfig, VariableType, VariableValue } from '@devcycle/types'
 import { RequestConsolidator } from './RequestConsolidator'
 import { dvcDefaultLogger } from './logger'
 import { DVCLogger } from '@devcycle/types'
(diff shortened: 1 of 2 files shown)


sdk/js/src/Client.ts
```diff
@@ -123,6 +123,11 @@ export class DVCClient implements Client {
                     ? EventTypes.variableDefaulted
                     : EventTypes.variableEvaluated,
                 target: variable.key,
+                metaData: {
+                    value: variable.value,
+                    type: typeof(variable.defaultValue),
+                    _variable: this.config?.variables?.[key]._id
```
kaushalkapasi, line 129: can probably use the existing data object defined on line 106?
```suggestion
                    _variable: data._id
```

sdk/js/src/Client.ts
```diff
@@ -123,6 +123,11 @@ export class DVCClient implements Client {
                     ? EventTypes.variableDefaulted
                     : EventTypes.variableEvaluated,
                 target: variable.key,
+                metaData: {
+                    value: variable.value,
+                    type: typeof(variable.defaultValue),
```
ajwootto, line 128: this should match the set of types defined here
https://github.com/taplytics/DevCycle/blob/main/lib/shared/mongo/schemas/src/lib/schemas/variable.ts#L120
ie. String, Number, Boolean or JSON

kaushalkapasi (approved):

sdk/js/src/Client.ts
```diff
@@ -118,11 +118,18 @@ export class DVCClient implements Client {
...
+            const variableFromConfig = this.config?.variables?.[key]
+
             this.eventQueue.queueAggregateEvent({
                 type: variable.isDefaulted
                     ? EventTypes.variableDefaulted
                     : EventTypes.variableEvaluated,
                 target: variable.key,
+                metaData: {
+                    value: variable.value,
+                    type: variableFromConfig?.type,
```
ajwootto, line 130: wouldn't this only work if the variable exists in the config? We wouldn't be sending the type if the variable used the default value

sdk/js/src/Client.ts
```diff
@@ -287,3 +295,18 @@ const checkIfEdgeEnabled = (
...
+    if (typeof defaultValue === 'boolean') {
+        return VariableType.boolean
+    } else if (typeof defaultValue === 'number') {
+        return VariableType.number
+    } else if (typeof defaultValue === 'string') {
+        return VariableType.string
+    } else if (typeof defaultValue === 'object') {
+        return VariableType.object
+    } else {
+        logger.debug('Default value is not of type Boolean, Number, String, or JSON')
```
ajwootto, line 309: is this log something that is meant to inform the user they're using the variable method wrong? If so, should it be a warning and include the name of the variable key being used? I imagine we might want it to log this one level higher (in the variable method) as part of the parameter validation

chungchantal: I've updated it to do `logger.warn` and to include the variable key in the message. As there's a ticket for addressing the mismatch in the variable method already, is this ok?

ajwootto (approved):

PR #241, opened 2022-08-24
Adding variableFeatureVariationMap to bucketed config

jonathannorris:


lib/shared/bucketing-assembly-script/__tests__/bucketing/bucketing.test.ts
```diff
@@ -163,6 +163,11 @@
             'featureVariationMap': {
                 '614ef6aa473928459060721a': '615357cf7e9ebdca58446ed0'
             },
+            'variableFeatureVariationMap': {
```
andrewtraviss, line 166: Let's just call this the `variableVariationMap`, since that's what it really is.

lib/shared/bucketing-assembly-script/__tests__/bucketing/bucketing.test.ts
```diff
@@ -227,6 +232,20 @@ describe('Config Parsing and Generating', () => {
                 '614ef6aa473928459060721a': '6153553b8cf4e45e0464268d',
                 '614ef6aa475928459060721a': '615382338424cb11646d7668'
             },
+            'variableFeatureVariationMap': {
+                'feature2.cool': {
```
andrewtraviss, line 236: Using the feature in the key like this seems a bit wrong, since the variable keys aren't feature-scoped. `feature1.cool` and `feature2.cool` are the same variable, just being evaluated within different features at different times.

Maybe this should be structured to use just the variable key as the top level key, and have a value of `${feature.id}.${variation.id}` that we use in the event schema?

jonathannorris: those are different variables though. And the aggregated event is generated from the varible key being pased into the variable method, so I think we have to use the `variable.key` as the main key.

I agree that we can use the `${feature.id}.${variation.id}` format for instead of the object. However we may still need it to be an array of `${feature.id}.${variation.id}` values as we may allow for a variable to be assigned in multiple features in the future?

andrewtraviss: If they were different variables they would be able to exist at the same time with different values. They can't, though. `cool` is globally-scoped and maps to a single value. I was under the impression that we never planned to do feature-scoped variables.

Even if we did allow a variable to be somehow related to multiple features in the future, we could support that by making the value a list of `${feature.id}.${variation.id}` elements.

andrewtraviss: Part of my thinking here is that we should minimize the number of distinct representations of this data, to make sure it's easy to correlate what we see happening in different places.

jonathannorris: I think you are getting confused here, that variable's key is literally just `feature2.cool`, other feature keys are named without the `.` in them.

andrewtraviss (approved):

PR #242, opened 2022-08-25
DVC-3131 Implementing EventQueue in AssemblyScript

jonathannorris:
Implement EventQueue code from NodeJS SDK into AssemblyScript so that the logic can be shared between SDKs. 
- See `eventQueueManager.ts` for main public interface of the EventQueue. 
- See `eventQueue.ts` for main user + aggregate event logic.
- See `requestPayloadManager.ts` for request payload generation logic + onSuccess / onFailure logic
- Added option to NodeJS SDK to have it use new `EventQueueAS.ts` file to use new Assembly Script logic.


sdk/nodejs/src/eventQueueAS.ts
```diff
@@ -0,0 +1,129 @@
...
+     */
+    async flushEvents(): Promise<void> {
+        const flushPayloadsStr = getBucketingLib().flushEventQueue(this.environmentKey)
+        this.logger.debug(`AS Flush Payloads: ${flushPayloadsStr}`)
+        const flushPayloads = JSON.parse(flushPayloadsStr) as FlushPayload[]
+        if (flushPayloads.length === 0) return
+
+        const innerReducer = (val: number, batch: UserEventsBatchRecord) => val + batch.events.length
+        const reducer = (val: number, batches: FlushPayload) => val + batches.records.reduce(innerReducer, 0)
+        const eventCount = flushPayloads.reduce(reducer, 0)
```
jonathannorris, line 67: this seems kinda expensive to do every time just for logs (copied this from nodeJS)...

lib/shared/bucketing-assembly-script/assembly/eventQueue/eventQueue.ts
```diff
@@ -0,0 +1,168 @@
...
+
+        let targetAggMap: Map<string, i64>
+        if (typeAggMap.has(target)) {
+            targetAggMap = typeAggMap.get(target)
+        } else {
+            targetAggMap = new Map<string, i64>()
+            typeAggMap.set(target, targetAggMap)
+        }
+
+        // TODO: Do we need to check that the type is aggVariableEvaluated here?
```
jonathannorris, line 144: @andrewtraviss thoughts?

lib/shared/bucketing-assembly-script/assembly/eventQueue/eventQueue.ts
```diff
@@ -0,0 +1,168 @@
...
+                const featureKey = targetVariations[i]
+                if (targetAggMap.has(featureKey)) {
+                    const featureCount: i64 = targetAggMap.get(featureKey)
+                    targetAggMap.set(featureKey, featureCount + 1)
+                } else {
+                    targetAggMap.set(featureKey, 1)
+                }
+            }
+        }
+        // TODO: Do we need to check that the type is aggVariableDefaulted here?
```
jonathannorris, line 158: @andrewtraviss thoughts?

sdk/nodejs/src/eventQueueAS.ts
andrewtraviss: Seems like the FlushPayload could at least have a field on it for the number of events in the payload, so that you wouldn't need this inner reducer? It could just be calculated when the FlushPayload is created.

Also, later on we check if logging is enabled or not. If logging is disabled, then we shouldn't be doing any of this either?

lib/shared/bucketing-assembly-script/assembly/eventQueue/requestPayloadManager.ts
```diff
@@ -0,0 +1,150 @@
...
+    markPayloadFailure(payloadId: string, retryable: boolean): void {
+        if (!this.pendingPayloads.has(payloadId)) {
+            throw new Error(
+                `Could not find payload: ${payloadId}, retryable: ${retryable} to mark as failure`
+            )
+        }
+
+        if (retryable) {
+            const payload = this.pendingPayloads.get(payloadId)
+            payload.status = 'failed'
```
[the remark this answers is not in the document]
jonathannorris: It doesn't currently. but probably should. there is just the `checkForFailedPayloads()` method currently.

lib/shared/bucketing-assembly-script/assembly/eventQueue/eventQueue.ts
jonathannorris: yea I guess the only thing is do we want to keep these methods more generic so that we could use them for other events in the future. Maybe I'll add some params to these methods to describe this functionality.

lib/shared/bucketing-assembly-script/assembly/types/dvcEvent.ts
```diff
@@ -0,0 +1,186 @@
...
+    stringify(): string {
+        const json = new JSON.Obj()
+        json.set('user', this.user)
+        json.set('events', jsonArrFromValueArray(this.events))
+        return json.stringify()
+    }
+}
+
+export class FlushPayload extends JSON.Value {
+    // TODO: implement
```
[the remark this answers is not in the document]
jonathannorris: will cleanup, this was for the `uuid()` implemented below.

sdk/nodejs/src/eventQueueAS.ts
```diff
@@ -0,0 +1,129 @@
+import { DVCEvent } from './types'
+import { DVCRequestEvent } from './models/requestEvent'
+import { DVCPopulatedUser } from './models/populatedUser'
+import { DVCLogger } from '@devcycle/types'
+
+import { getBucketingLib } from './bucketing'
+import { EventQueueInterface } from './eventQueue'
+import { publishEvents } from './request'
+
+export const AggregateEventTypes: Record<string, string> = {
```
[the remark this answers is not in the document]
jonathannorris: yup, will fix.

sdk/nodejs/src/eventQueueAS.ts
jonathannorris: yea makes sense

lib/shared/bucketing-assembly-script/assembly/eventQueue/eventQueue.ts
```diff
@@ -0,0 +1,168 @@
...
+    queueEvent(user: DVCPopulatedUser, event: DVCEvent, bucketedConfig: BucketedUserConfig): void {
+        // TODO: Implement max queue size
+        // this.maxEventQueueSize = bucketedConfig?.project.settings.sdkSettings?.eventQueueLimit ?? this.maxEventQueueSize
+
+        const requestEvent = new DVCRequestEvent(event, user.user_id, bucketedConfig.featureVariationMap)
+        this.addEventToQueue(user, requestEvent)
+    }
+
+    private addEventToQueue(user: DVCPopulatedUser, event: DVCRequestEvent): void {
+        // TODO: implement?
```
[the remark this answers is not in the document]
jonathannorris: The other option is to have the WASM return a known Error type and have the Native code handle that. I've made implementing this its own ticket: https://taplytics.atlassian.net/browse/DVC-4285

lib/shared/bucketing-assembly-script/assembly/managers/eventQueueManager.ts
```diff
@@ -0,0 +1,64 @@
...
+    const eventQueue = getEventQueue(envKey)
+    eventQueue.onPayloadFailure(payloadId, retryable)
+}
+
+export function queueEvent(envKey: string, userStr: string, eventStr: string): void {
+    const eventQueue = getEventQueue(envKey)
+    const dvcUser = DVCPopulatedUser.fromJSONString(userStr)
+    const event = DVCEvent.fromJSONString(eventStr)
+
+    const bucketedConfig = _generateBucketedConfig(_getConfigData(envKey), dvcUser)
```
jonathannorris, line 53: hum... @andrewtraviss I think we are going to need to change this method to pass in the bucketed config it used when evaluating the variables for these events, as its possible that the bucketed user config no-longer matches the one used when evaluating the variable...

sdk/nodejs/src/eventQueueAS.ts
```diff
@@ -0,0 +1,129 @@
...
+    events: DVCRequestEvent[]
+}
+export type FlushPayload = {
+    payloadId: string,
+    eventCount: number,
+    records: UserEventsBatchRecord[]
+}
+
+type options = {
+    flushEventsMS?: number,
```
laurawarr, line 21: this param name wasn't obvious that it's the interval, what about something like `eventFlushInterval`?

sdk/nodejs/src/eventQueueAS.ts
```diff
@@ -0,0 +1,129 @@
...
+        if (!flushPayloadsStr) return
+        this.logger.debug(`AS Flush Payloads: ${flushPayloadsStr}`)
+        const flushPayloads = JSON.parse(flushPayloadsStr) as FlushPayload[]
+        if (flushPayloads.length === 0) return
+
+        const reducer = (val: number, batches: FlushPayload) => val + batches.eventCount
+        const eventCount = flushPayloads.reduce(reducer, 0)
+        this.logger.debug(`DVC Flush ${eventCount} AS Events, for ${flushPayloads.length} Users`)
+
+        await Promise.all(flushPayloads.map(async (flushPayload) => {
```
laurawarr, line 70: any concern about sending these requests all at once? should we consider sending the catch requests in batches?

lib/shared/bucketing-assembly-script/assembly/eventQueue/eventQueue.ts
```diff
@@ -0,0 +1,197 @@
...
+    EventQueueOptions,
+    DVCRequestEvent,
+    UserEventsBatchRecord,
+    FlushPayload,
+    FeatureVariation
+} from '../types'
+import { jsonArrFromValueArray } from '../helpers/jsonHelpers'
+import { RequestPayloadManager } from './requestPayloadManager'
+
+export type AggEventQueue = Map<string, Map<string, Map<string, Map<string, i64>>>>
```
ajwootto, line 13: it might help understanding to unpack this into multiple sub types that are named descriptively for each level of nesting

lib/shared/bucketing-assembly-script/__tests__/types/dvcEvent.test.ts
```diff
@@ -0,0 +1,75 @@
...
+        JSON.stringify(event),
+        user_id as string,
+        JSON.stringify(featureVars)
+    ))
+}
+
+describe('DVCEvent tests', () => {
+    describe('DVCEvent', () => {
+        it('should test DVCEvent json parsing', () => {
+            const eventObj = {
```
ajwootto, line 21: should you be testing for invalid values here or something

lib/shared/bucketing-assembly-script/assembly/eventQueue/eventQueue.ts
```diff
@@ -0,0 +1,197 @@
...
+
+        // TODO: implement?
+        // if (this.eventQueueSize() >= this.maxEventQueueSize) {
+        //     this.logger.warn(`Max event queue size reached, dropping aggregate event: ${event}`)
+        //     this.flushEvents()
+        //     return
+        // }
+
+        let variableFeatureVarAggMap: Map<string, Map<string, Map<string, i64>>>
+        if (this.aggEventQueue.has(type)) {
```
ajwootto, line 142: might be nice to break this method up into more manageable steps implemented as private functions, like:
`ensureHasType`, `ensureHasTarget`, `aggregateByVariation` etc

lib/shared/bucketing-assembly-script/assembly/eventQueue/requestPayloadManager.ts
```diff
@@ -0,0 +1,173 @@
...
+
+    constructFlushPayloads(
+        userEventQueue: Map<string, UserEventsBatchRecord>,
+        aggEventQueue: AggEventQueue
+    ): FlushPayload[] {
+        this.checkForFailedPayloads()
+
+        // Add events from user event queue
+        const userEventQueueKeys = userEventQueue.keys()
+        for (let i = 0; i < userEventQueueKeys.length; i++) {
```
ajwootto, line 34: can't you just use `Object.entries` here?

ajwootto: oh nvm its not a plain object

lib/shared/bucketing-assembly-script/assembly/eventQueue/requestPayloadManager.ts
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    /**
+     * Check that the pendingPayloads queue is empty or only contains retryable failed payloads
+     * before creating new batch of payloads.
+     */
+    checkForFailedPayloads(): void {
+        this.pendingPayloads.values().forEach((payload) => {
+            if (payload.status === 'failed') {
+                payload.status = 'sending'
```
ajwootto, line 167: won't this status change be incorrect if the code encounters a failed payload first, but then ends up throwing an error due to a different payload in the queue later on? Then you'll have changed some of the failed payloads to "sending" before throwing the error.

lib/shared/bucketing-assembly-script/assembly/eventQueue/requestPayloadManager.ts
```diff
@@ -0,0 +1,173 @@
...
+        this.addEventsToPendingPayloads(new UserEventsBatchRecord(dvcUser, aggEvents))
+    }
+
+    /**
+     * Chunk up UserEventsBatchRecord's into payloads of size: this.chunkSize
+     */
+    private addEventsToPendingPayloads(record: UserEventsBatchRecord): void {
+        let flushPayload = new FlushPayload([])
+
+        while (record.events.length > this.chunkSize) {
```
ajwootto, line 117: why not just make this condition `record.events.length > 0`?

lib/shared/bucketing-assembly-script/assembly/eventQueue/requestPayloadManager.ts
```diff
@@ -0,0 +1,173 @@
...
+            const batchRecord = new UserEventsBatchRecord(
+                record.user,
+                record.events.splice(0, this.chunkSize)
```
ajwootto, line 120: it seems a bit weird to directly mutate the records that are being passed in here

lib/shared/bucketing-assembly-script/assembly/eventQueue/requestPayloadManager.ts
```diff
@@ -0,0 +1,173 @@
...
+        }
+
+        // Add events from aggregate events
+        this.addAggEventsToPendingPayloads(aggEventQueue)
+
+        return this.pendingPayloads.values()
+    }
+
+    /**
+     * generate aggregated events by resolving aggregated event map into DVCEvent's.
```
ajwootto, line 47:
```suggestion
     * generate aggregated events by resolving aggregated event map into DVCEvents.
```

lib/shared/bucketing-assembly-script/assembly/eventQueue/requestPayloadManager.ts
```diff
@@ -0,0 +1,173 @@
...
+     */
+    private addAggEventsToPendingPayloads(
+        aggEventQueue: AggEventQueue
+    ): void {
+        const aggEventQueueKeys = aggEventQueue.keys()
```
ajwootto, line 52: it would be a lot easier to understand this part if the various `keys` arrays were named after what they actually contain, like `featureIds` and whatnot

lib/shared/bucketing-assembly-script/assembly/types/dvcEvent.ts
```diff
@@ -0,0 +1,213 @@
...
+    getStringFromJSON,
+    getStringFromJSONOptional,
+    getStringMapFromJSONObj,
+    jsonArrFromValueArray,
+    jsonObjFromMap
+} from '../helpers/jsonHelpers'
+import { DVCPopulatedUser } from './dvcUser'
+import uuid from 'as-uuid/assembly'
+
+interface DVCUserInterface {
```
ajwootto, line 15: DVCEventInterface?

lib/shared/bucketing-assembly-script/assembly/types/dvcEvent.ts
```diff
@@ -0,0 +1,213 @@
...
+        const date = this.date // nullable check only works like this
+        if (date) json.set('date', date.toISOString())
+        if (this.target) json.set('target', this.target)
+        if (!isNaN(this.value)) json.set('value', this.value)
+        if (this.metaData) json.set('metaData', this.metaData)
+        return json.stringify()
+    }
+}
+
+export class DVCRequestEvent extends JSON.Value {
```
ajwootto, line 85: perhaps document in a comment what this class is for and how its different from DVCEvent

lib/shared/bucketing-assembly-script/assembly/eventQueue/eventQueue.ts
```diff
@@ -0,0 +1,197 @@
...
+        // TODO: implement?
+        // if (this.eventQueueSize() >= this.maxEventQueueSize) {
+        //     this.logger.warn(`Max event queue size reached, dropping event: ${event}`)
+        //     this.flushEvents()
+        //     return
+        // }
+
+        const user_id = user.user_id
+        let userEvents: UserEventsBatchRecord
+        if (!this.userEventQueue.has(user_id)) {
```
ajwootto, line 113: grouping the events purely by user_id means that you'll be using the user data as specified by the first event that was queued, even if the user data changes later on (but keeps the same user id). Shouldn't you group them by the whole user data object to make sure the data is correct on each event?

lib/shared/bucketing-assembly-script/assembly/types/bucketedUserConfig.ts
```diff
@@ -109,15 +148,15 @@ export class SDKFeature extends JSON.Obj {
         super()
     }
 
-    static sdkFeatureFromJSONObj(feature: JSON.Obj): SDKFeature {
+    static fromJSONString(featureStr: JSON.Obj): SDKFeature {
```
ajwootto, line 151: seems a bit weird to call a JSON.Obj variable `featureStr`

ajwootto: this is technically `fromJSONObj` is it not?

sdk/nodejs/src/eventQueueAS.ts
```diff
@@ -0,0 +1,129 @@
...
+                    } else {
+                        getBucketingLib().onPayloadFailure(this.environmentKey, flushPayload.payloadId, false)
+                    }
+                } else {
+                    this.logger.debug(`DVC Flushed ${eventCount} Events, for ${flushPayload.records.length} Users`)
+                    getBucketingLib().onPayloadSuccess(this.environmentKey, flushPayload.payloadId)
+                }
+            } catch (ex) {
+                this.logger.error(`DVC Error Flushing Events response message: ${ex.message}`)
+                getBucketingLib().onPayloadFailure(this.environmentKey, flushPayload.payloadId, false)
```
ajwootto, line 86: this one is retryable

lib/shared/bucketing-assembly-script/assembly/types/dvcUser.ts
```diff
@@ -18,48 +18,49 @@ interface DVCUserInterface {
...
         const privateCustomData = user.getObj('privateCustomData')
         if (!isFlatJSONObj(privateCustomData)) {
             throw new Error('DVCUser privateCustomData can\'t contain nested objects or arrays')
         }
-        this.privateCustomData = privateCustomData
 
-        return this
+        return new DVCUser(
+            getStringFromJSON(user, 'user_id'),
+            // Need to set a default "null" value, as numbers can't be null in AS
```
ajwootto, line 53: what is this comment in reference to?

sdk/nodejs/src/eventQueueAS.ts
jonathannorris: I don't have too many concerns about that, seems like an early optimzation for me. with the batch size of 100 and max size of 1000, this would be at most 10 requests.

lib/shared/bucketing-assembly-script/assembly/eventQueue/eventQueue.ts
jonathannorris: the user is updated to the latest user in line `106` below

lib/shared/bucketing-assembly-script/__tests__/types/dvcEvent.test.ts
jonathannorris: That is already handled in other test files.

sdk/nodejs/__tests__/cloudClient.spec.ts
```diff
@@ -115,22 +115,22 @@ describe('DVCCloudClient without EdgeDB', () => {
         it('throws an error for a null event', () => {
             const badEvent = null
             expect(async () => await client.track(user, badEvent as unknown as DVCEvent))
```
andrewtraviss, line 117: I don't think the async and await are needed in these tests. The underlying function returns a promise, which `expect()` can handle directly

sdk/nodejs/__tests__/eventQueueAS.test.ts
```diff
@@ -0,0 +1,660 @@
...
+                    user_id: 'host.name',
+                    createdDate: expect.any(String),
+                    lastSeenDate: expect.any(String),
+                    platform: 'NodeJS',
+                    platformVersion: expect.any(String),
+                    sdkType: 'server',
+                    sdkVersion: expect.any(String)
+                }),
+                events: [
+                    expect.objectContaining({
```
andrewtraviss, line 129: The tests would be a lot more succinct and easier to understand if we had a function `expectEvent` which you could pass an object into with only the fields the test actually cares about the values of. That function would `expect.any()` unspecified fields as appropriate.

andrewtraviss (approved):

PR #244, opened 2022-09-02
Fix nodejs sdk startup without options

kaushalkapasi:
- make `options` optional when calling `initialize` on the NodeJS SDK, this makes the NodeJS SDK startup in Local bucketing mode by default
- update NodeJS example apps

### DIFF
--- a/sdk/nodejs/__tests__/initialize.spec.ts
+++ b/sdk/nodejs/__tests__/initialize.spec.ts
@@ -1,0 +1,32 @@
+import { DVCClient, DVCCloudClient, initialize } from '../src/index'
+
+jest.mock('../src/bucketing')
+jest.mock('../src/environmentConfigManager')
+
+describe('NodeJS SDK Initialize', () => {
+    afterAll(() => {
+        jest.clearAllMocks()
+    })
+
+    it('sucessfully calls initialize with no options', async () => {
+        const client: DVCClient = await initialize('token').onClientInitialized()
+        expect(client).toBeDefined()
+    })
+
+    it('fails to initialize in Local Bucketing mode when no token is provided', () => {
+        expect(() => 
+            initialize(undefined as unknown as string)
+        ).toThrow('Missing environment key! Call initialize with a valid environment key')
+    })
+
+    it('sucessfully calls initialize with enableCloudBucketing set to true', () => {
+        const client: DVCCloudClient = initialize('token', { enableCloudBucketing: true })
+        expect(client).toBeDefined()
+    })
+
+    it('fails to initialize in Cloud Bucketing mode when no token is provided', () => {
+        expect(() =>
+            initialize(undefined as unknown as string, { enableCloudBucketing: true })
+        ).toThrow('Missing environment key! Call initialize with a valid environment key')
+    })
+})
--- a/sdk/nodejs/src/index.ts
+++ b/sdk/nodejs/src/index.ts
@@ -8,14 +8,14 @@ export * from './types'
 type DVCOptionsCloudEnabled = DVCOptions & { enableCloudBucketing: true }
 type DVCOptionsLocalEnabled = DVCOptions & { enableCloudBucketing?: false }
 
-export function initialize(environmentKey: string, options: DVCOptionsLocalEnabled): DVCClient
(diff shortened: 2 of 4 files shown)


jonathannorris (approved):

jsalaber (approved):

sdk/nodejs/src/index.ts
```diff
@@ -8,14 +8,14 @@ export * from './types'
 type DVCOptionsCloudEnabled = DVCOptions & { enableCloudBucketing: true }
 type DVCOptionsLocalEnabled = DVCOptions & { enableCloudBucketing?: false }
 
-export function initialize(environmentKey: string, options: DVCOptionsLocalEnabled): DVCClient
+export function initialize(environmentKey: string, options?: DVCOptionsLocalEnabled): DVCClient
 export function initialize(environmentKey: string, options: DVCOptionsCloudEnabled): DVCCloudClient
-export function initialize(environmentKey: string, options: DVCOptions): DVCClient | DVCCloudClient {
+export function initialize(environmentKey: string, options?: DVCOptions): DVCClient | DVCCloudClient {
```
ajwootto, line 13: just default options to an empty object here

ajwootto (changes requested):

sdk/nodejs/src/index.ts
ajwootto: otherwise you gotta worry about null checking it all throughout this method

ajwootto, line 13: there should be a test for calling this with no options

sdk/nodejs/__tests__/initialize.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+
+    it('sucessfully calls initialize with no options', async () => {
+        const client: DVCClient = await initialize('token').onClientInitialized()
+        expect(client).toBeDefined()
+    })
+
+    it('fails to initialize in Local Bucketing mode when no token is provided', async () => {
+        try {
+            await initialize(undefined as unknown as string)
+            // Fail test if above expression doesn't throw anything.
```
ajwootto, line 19: you can also write this as :
`await expect(() => initialize(undefined as unknown as string)).rejects.toThrow('Missing environment key! Call initialize with a valid environment key')`

kaushalkapasi: Ah! Thanks.. that's what I couldn't figure out how to write correctly.. that's why I went with the try/catch format.. will revise it! thanks 🙏

sdk/nodejs/__tests__/initialize.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+        jest.clearAllMocks()
+    })
+
+    it('sucessfully calls initialize with no options', async () => {
+        const client: DVCClient = await initialize('token').onClientInitialized()
+        expect(client).toBeDefined()
+    })
+
+    it('fails to initialize in Local Bucketing mode when no token is provided', async () => {
+        await expect(async () =>
```
ajwootto, line 17: dont need the extra async await here

kaushalkapasi: test failed without it? 🤔

ajwootto: they shouldn't, its exactly equivalent to just write a plain method that returns the promise

kaushalkapasi: we don't need the async/await outside the expect, just inside.. cuz this way, it awaits the initialize, then proceeds to check the reject.. :)

ajwootto: that's not true either, if you dont await the expect then the test will pass no matter what

ajwootto: I think you can actually even drop the inner function on the expect and just call initialize directly, which then passes expect the promise

kaushalkapasi: 🤦 not an async function.. so don't need any async/await / rejects

ajwootto (approved):

PR #247, opened 2022-09-06
DVC-4333 update format of aggVariableEvaluated events

jonathannorris:


lib/shared/bucketing-assembly-script/assembly/eventQueue/eventQueue.ts
```diff
@@ -21,20 +23,30 @@ export class EventQueue {
     private userEventQueue: Map<string, UserEventsBatchRecord>
 
     /**
-     * Map<'aggVariableDefaulted' | 'aggVariableEvaluated,
+     * Map<'aggVariableDefaulted',
      *      Map<variable.key,
-     *          Map<`${feature_id}.${variation_id}`, counter>
+     *          Map<'value',
+     *              Map<'value', counter>
+     *          >
```
laurawarr, line 30: what does the second nested value represent?

laurawarr, line 30: what does the second value map represent?

jonathannorris: just basically have to have the aggVariableDefaulted events have a map like: `{value: {value: counter}}` to fit into the same map: https://github.com/DevCycleHQ/js-sdks/pull/247/files/007f1e366e5450623234dd06cc09a881be1084fd#diff-dd534d11622b593c587e8002c56f9c20e6349aee998e52c50c580580a1486ce9R13

We could split those events into their own top-level map if you think that's better.

laurawarr (approved):

PR #248, opened 2022-09-07
DVC-4287 NodeJS SDK EventQueueAS tests, restructure requestPayloadManager

jonathannorris:
- Copy / update existing NodeJS SDK EventQueue tests to be run against EventQueueAS and AS code.
- Restructure requestPayloadManager to produce payloads with multiple users' data in the payloads.


lib/shared/bucketing-assembly-script/assembly/types/eventQueueOptions.ts
```diff
@@ -1,13 +1,7 @@
 import { JSON } from 'assemblyscript-json/assembly'
-import {
-    getF64FromJSONValue,
-    getI32FromJSONValue
-} from '../helpers/jsonHelpers'
+import { getI32FromJSONValue } from '../helpers/jsonHelpers'
 
 export class EventQueueOptions extends JSON.Obj {
-    flushEventsMS: f64 = 10 * 1000
```
jonathannorris, line 8: none of these options were used in AS.

sdk/nodejs/__tests__/eventQueueAS.test.ts
```diff
@@ -0,0 +1,594 @@
...
+        ).toBe(100)
+
+        expect(
+            fourthPublishPayloads.reduce((val: number, payload) => val + payload.events.length, 0)
+        ).toBe(100)
+
+        eventQueue.cleanup()
+    })
+
+    // TODO: implement when this is implemented
```
jonathannorris, line 496: I can delete these if we want, but going to do this PR next.

sdk/nodejs/project.json
```diff
@@ -21,19 +21,20 @@
...
-      "outputs": [
-        "coverage/sdk/nodejs"
-      ],
+      "executor": "@nrwl/workspace:run-commands",
       "options": {
-        "jestConfig": "sdk/nodejs/jest.config.ts",
-        "passWithNoTests": true
+          "commands": [
+              {
+                  "command": "node --experimental-vm-modules ../../node_modules/jest/bin/jest.js"
```
jonathannorris, line 28: Had to switch to this executor to allow the WASM to be executed by jest.

sdk/nodejs/src/eventQueueAS.ts
```diff
@@ -65,6 +65,7 @@ export class EventQueueAS implements EventQueueInterface {
 
         const reducer = (val: number, batches: FlushPayload) => val + batches.eventCount
         const eventCount = flushPayloads.reduce(reducer, 0)
+        console.log(`DVC Flush ${eventCount} AS Events, for ${flushPayloads.length} Users`)
```
ajwootto, line 68: log

sdk/nodejs/src/eventQueueAS.ts
```diff
@@ -40,7 +40,7 @@ export class EventQueueAS implements EventQueueInterface {
 
         this.flushInterval = setInterval(this.flushEvents.bind(this), this.flushEventsMS)
 
-        getBucketingLib().initEventQueue(environmentKey, JSON.stringify(options))
+        getBucketingLib().initEventQueue(environmentKey, JSON.stringify(options || {}))
```
ajwootto, line 43: probably just default the options param to an empty object?

jonathannorris: yea duh, can't do that in AS, so that's where my brain was.

sdk/nodejs/project.json
jonathannorris: Is the same way we run in the `bucketing-as` lib

lib/shared/bucketing-assembly-script/assembly/types/eventQueueOptions.ts
ajwootto: shouldn't we be handling the disabling of event logging in AS?

lib/shared/bucketing-assembly-script/assembly/types/dvcEvent.ts
```diff
@@ -139,8 +139,8 @@ export class DVCRequestEvent extends JSON.Value {
         if (this.target) json.set('target', this.target)
         if (this.customType) json.set('customType', this.customType)
         json.set('user_id', this.user_id)
-        json.set('date', this.date)
-        json.set('clientDate', this.clientDate)
+        json.set('date', this.date.toISOString())
```
ajwootto, line 142: why the switch to date strings?

lib/shared/bucketing-assembly-script/assembly/eventQueue/requestPayloadManager.ts
```diff
@@ -101,32 +101,59 @@ export class RequestPayloadManager {
...
+            flushPayload.addBatchRecordForUser(record, this.chunkSize)
             this.pendingPayloads.set(flushPayload.payloadId, flushPayload)
             flushPayload = new FlushPayload([])
         }
         if (record.events.length > 0) {
-            flushPayload.records.push(record)
+            flushPayload.addBatchRecordForUser(record, this.chunkSize)
         }
-        if (flushPayload.records.length > 0) {
+        if (flushPayload.eventCount() > 0) {
```
ajwootto, line 156: this is kind of confusing, are you potentially constructing flushPayload objects that dont end up getting used and therefore need to check if they have any events before adding to the pendingPayloads map?

lib/shared/bucketing-assembly-script/assembly/eventQueue/requestPayloadManager.ts
```diff
@@ -101,32 +101,59 @@ export class RequestPayloadManager {
...
-        let flushPayload = new FlushPayload([])
+        let flushPayload = this.getFlushPayloadForUser_id(record.user.user_id)
 
-        while (record.events.length > this.chunkSize) {
-            const batchRecord = new UserEventsBatchRecord(
-                record.user,
-                record.events.splice(0, this.chunkSize)
-            )
-            flushPayload.records.push(batchRecord)
+        while (record.events.length > (this.chunkSize - flushPayload.eventCount())) {
```
ajwootto, line 148: it would be clearer to me if this just looped until record.events was empty (or some similar condition asuming you stop mutating that array) and then the contents of the loop would have an `if-else` check for whether to use an existing payload or create a new one

lib/shared/bucketing-assembly-script/assembly/eventQueue/requestPayloadManager.ts
```diff
@@ -101,32 +101,59 @@ export class RequestPayloadManager {
...
+            return new FlushPayload([])
+        }
+    }
+
     /**
      * Chunk up UserEventsBatchRecord's into payloads of size: this.chunkSize
      */
     private addEventsToPendingPayloads(record: UserEventsBatchRecord): void {
-        let flushPayload = new FlushPayload([])
+        let flushPayload = this.getFlushPayloadForUser_id(record.user.user_id)
```
ajwootto, line 146: personally I think the flush payloads should be immutable once created. What you're doing here potentially adds more events to a payload that was already sent and then failed. Once a payload is returned to native code I dont think it should be possible for it to change.

lib/shared/bucketing-assembly-script/assembly/eventQueue/requestPayloadManager.ts
```diff
@@ -101,32 +101,59 @@ export class RequestPayloadManager {
...
+        const keys = this.pendingPayloads.keys()
+        let smallestPayload: FlushPayload | null = null
+
+        for (let i=0; i < keys.length; i++) {
+            const key = keys[i]
+            const payload = this.pendingPayloads.get(key)
+            const recordForUser = payload.recordForUser(user_id)
+            const payloadEventCount = payload.eventCount()
+            if (recordForUser && payloadEventCount < this.chunkSize) {
+                this.pendingPayloads.delete(key)
```
ajwootto, line 125: why does the found payload need to be deleted from pending payloads before being returned? Is it because you're then going to add more stuff to it? Maybe another good reason they should be immutable?

lib/shared/bucketing-assembly-script/assembly/types/dvcEvent.ts
jonathannorris: that's what the API docs have: https://events.devcycle.com/docs#/Events/EventsController_submitEventsBatchV1

lib/shared/bucketing-assembly-script/assembly/eventQueue/requestPayloadManager.ts
jonathannorris: yea I can make an exception for `failed` payloads if we want, however the reason this is needed is because user events and aggregated events are added to the payloads seperatly. Which is what was resulting in the system creating two payloads every time you logged a user event an aggVariableAggregated: https://github.com/DevCycleHQ/js-sdks/pull/248/files/79a82abe111f7aa19e7c1c58b8691c4acee419c6#diff-5e909c6d5ed133d11580490d906b2f23a686d4e7e83e34ddac71713eb3100dc3R26

lib/shared/bucketing-assembly-script/assembly/eventQueue/requestPayloadManager.ts
jonathannorris: yea that makes sense.

lib/shared/bucketing-assembly-script/assembly/eventQueue/requestPayloadManager.ts
```diff
@@ -101,32 +101,59 @@ export class RequestPayloadManager {
...
+        let flushPayload = this.getFlushPayloadForUser_id(record.user.user_id)
 
-        while (record.events.length > this.chunkSize) {
-            const batchRecord = new UserEventsBatchRecord(
-                record.user,
-                record.events.splice(0, this.chunkSize)
-            )
-            flushPayload.records.push(batchRecord)
+        while (record.events.length > (this.chunkSize - flushPayload.eventCount())) {
+            flushPayload.addBatchRecordForUser(record, this.chunkSize)
```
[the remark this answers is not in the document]
jonathannorris: I don't see how we can avoid that if we want to ensure that paylaods are always full with their maximum number of events, across multiple users.

lib/shared/bucketing-assembly-script/assembly/types/eventQueueOptions.ts
jonathannorris: it can be, currently its handled on the Native side.

jonathannorris: https://taplytics.atlassian.net/browse/DVC-4408

lib/shared/bucketing-assembly-script/assembly/eventQueue/requestPayloadManager.ts
jonathannorris: yea basically because the calling code always adds it. I'll try and restructure this to remove this delete.

lib/shared/bucketing-assembly-script/assembly/eventQueue/requestPayloadManager.ts
```diff
@@ -28,17 +33,19 @@ export class RequestPayloadManager {
...
         for (let i = 0; i < userEventQueueKeys.length; i++) {
             const key = userEventQueueKeys[i]
             const userEventsRecord = userEventQueue.get(key)
-            this.addEventsToPendingPayloads(userEventsRecord)
+            records.push(userEventsRecord)
         }
 
         // Add events from aggregate events
-        this.addAggEventsToPendingPayloads(aggEventQueue)
+        records.push(this.addAggEventsToPendingPayloads(aggEventQueue))
```
ajwootto, line 47: you should probably rename this method now that its purpose has changed

`constructBatchRecordsFromAggEvents` or something

lib/shared/bucketing-assembly-script/assembly/eventQueue/requestPayloadManager.ts
```diff
@@ -101,33 +108,46 @@ export class RequestPayloadManager {
...
+        for (let i = 0; i < records.length; i++) {
+            const record: UserEventsBatchRecord = records[i]
+
+            do {
+                const flushPayload = this.getFlushPayload()
+                flushPayload.addBatchRecordForUser(record, this.chunkSize)
+                if (flushPayload.records.length > 0) {
+                    this.pendingPayloads.set(flushPayload.payloadId, flushPayload)
+                }
+            } while (record.events.length > 0)
```
ajwootto, line 150: I think you still want this to be a regular while loop, there's no reason to run it even once if there are not events in the record

lib/shared/bucketing-assembly-script/__tests__/eventQueue/eventQueueManager.test.ts
```diff
@@ -44,7 +46,12 @@ const initSDK = (envKey: string, eventOptions: unknown = {}) => {
 }
 
 describe('EventQueueManager Tests', () => {
-    afterEach(() => clearPlatformData())
+    afterEach(() => {
+        clearPlatformData()
+        if (currentEnvKey) {
+            cleanupEventQueue(currentEnvKey)
+        }
```
andrewtraviss, line 53: Should `currentEnvKey` also be cleared here?

jonathannorris: it can be yea.

andrewtraviss (approved):

PR #254, opened 2022-09-14
add git push to lerna script

L-Joli:


scripts/lerna-version.sh
```diff
@@ -70,3 +70,5 @@ git commit --amend --no-edit --no-verify
 while IFS= read -r line; do
   git tag -f "$line" -m "$line"
 done <<< "$RELEASE_TAGS"
+
+git push && git push --tags
```
kaushalkapasi, line 74: can you update the tags push to `git push --follow-tags` 
```suggestion
git push && git push --follow-tags
```

kaushalkapasi: also, do we want `git push` here?

L-Joli: I don't think we need it now [https://github.com/git/git/blob/master/Documentation/RelNotes/1.8.3.txt#L155-L156](https://github.com/git/git/blob/master/Documentation/RelNotes/1.8.3.txt#L155-L156)

ajwootto (approved):

PR #278, opened 2022-09-27
use fetch API and node-fetch (fallback) for NodeJS SDK

ajwootto:
- use cross-fetch to implement a fetch-compatible request layer in NodeJS SDK
- if global fetch is available, use that instead
- this will hopefully allow the SDK to be used in Cloudflare workers while taking advantage of native Fetch caching


ajwootto:
* main
  * **PR #278** <a href="https://app.graphite.dev/github/pr/DevCycleHQ/js-sdks/278" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/DevCycleHQ/js-sdks/278?utm_source=stack-comment).

sdk/nodejs/src/environmentConfigManager.ts
```diff
@@ -79,10 +78,11 @@ export class EnvironmentConfigManager {
 
         if (res?.status === 304) {
             this.logger.debug(`Config not modified, using cache, etag: ${this.configEtag}`)
-        } else if (res?.status === 200 && res?.data) {
-            getBucketingLib().setConfigData(this.environmentKey, JSON.stringify(res.data))
+        } else if (res?.status === 200) {
+            const projectConfig = await res.json()
+            getBucketingLib().setConfigData(this.environmentKey, JSON.stringify(projectConfig))
```
jonathannorris, line 83: what if `projectConfig` is null here?

sdk/nodejs/src/request.ts
```diff
@@ -43,86 +14,107 @@ const TRACK_PATH = '/v1/track'
...
+    status: number
+}
+
+const exponentialBackoff: RequestInitWithRetry['retryDelay'] = (
+    attempt, error, response
+) => {
+    return Math.pow(2, attempt) * 1000
+}
+
+const retryOnRequestError: (retries: number) => RequestInitWithRetry['retryOn'] = (retries) => (
```
jonathannorris, line 27: this double function definition is confusing, any way to split this up to make it more clear what's happening?

sdk/nodejs/src/request.ts
```diff
@@ -135,39 +127,74 @@ export async function postTrack(
...
 }
 
-export async function get(
-    requestConfig: AxiosRequestConfig,
-    retryable = true
-): Promise<AxiosResponse> {
-    const client = retryable ? axiosRetryableClient : axiosClient
-    return await client.request({
+export async function get(url: string, requestConfig: RequestInit | RequestInitWithRetry
+): Promise<Response> {
```
jonathannorris, line 157:
```suggestion
export async function get(
    url: string, 
    requestConfig: RequestInit | RequestInitWithRetry
): Promise<Response> {
```

sdk/nodejs/src/request.ts
```diff
@@ -135,39 +127,74 @@ export async function postTrack(
...
 
-export async function post(
-    requestConfig: AxiosRequestConfig,
-    envKey: string,
-    retryable = true
-): Promise<AxiosResponse> {
-    const client = retryable ? axiosRetryableClient : axiosClient
-    return await client.request({
+export async function post(url: string, requestConfig: RequestInit | RequestInitWithRetry, envKey: string
+): Promise<Response> {
```
jonathannorris, line 144:
```suggestion
export async function post(
    url: string, 
    requestConfig: RequestInit | RequestInitWithRetry, 
    envKey: string
): Promise<Response> {
```

sdk/nodejs/src/client.ts
```diff
@@ -180,7 +180,9 @@ export class DVCClient {
     }
 
     async close(): Promise<void> {
+        console.log('CLOSE CALLED')
```
jonathannorris, line 183: cleanup
```suggestion
```

sdk/nodejs/src/client.ts
```diff
@@ -180,7 +180,9 @@ export class DVCClient {
...
         await this.onInitialized
+        console.log('CLEANING UP')
```
jonathannorris, line 185: cleanup
```suggestion
```

jonathannorris (approved):

sdk/nodejs/src/cloudClient.ts
```diff
@@ -22,13 +21,34 @@ const castIncomingUser = (user: DVCUser) => {
...
+        if (code !== 404 && code < 500) {
+            throw new Error(`DevCycle request failed with status ${code}. ${err.message || ''}`)
+        }
+
+        // Catch non-4xx errors so we can log them instead
+        return
     }
+
+    if (err instanceof SyntaxError) {
+        // JSON parse error, log instead of throwing
```
jonathannorris, line 46: where does it get logged?

ajwootto: in the catch statements that are calling this. This method was a new thing we added to throw on 4xx errors, before this we just didn't have this method so the errors would always get caught and logged

sdk/nodejs/src/cloudClient.ts
```diff
@@ -65,13 +86,15 @@ export class DVCCloudClient {
                     })
                 }
                 return new DVCVariable({
-                    ...variable,
+                    ...variableResponse,
                     defaultValue
                 })
             })
-            .catch((err: AxiosError) => {
+            .catch((err: unknown) => {
```
jonathannorris, line 93: can these be `Error` instead of `unknown`?

ajwootto: no, because any "catch" statement in JS can potentially catch "anything". Code can throw whatever it wants, it doesn't have to be an Error. Declaring that what it's going to catch is definitely an "Error" isn't really correct

sdk/nodejs/src/environmentConfigManager.ts
```diff
@@ -80,10 +81,10 @@ export class EnvironmentConfigManager {
 
         if (res?.status === 304) {
             this.logger.debug(`Config not modified, using cache, etag: ${this.configEtag}`)
-        } else if (res?.status === 200 && res?.data) {
-            getBucketingLib().setConfigData(this.environmentKey, JSON.stringify(res.data))
+        } else if (res?.status === 200 && projectConfig) {
+            getBucketingLib().setConfigData(this.environmentKey, JSON.stringify(projectConfig))
```
jonathannorris, line 85: Optimization here, can we not call `res.json()` and just pass in `res.text()`?

ajwootto: technically yes but currently we rely on the JSON parsing exception that gets thrown by parsing this response. If you pass a non-parseable string into the bucketing lib it's a different error that's a bit less specific

sdk/nodejs/src/request.ts
```diff
@@ -43,87 +14,119 @@ const TRACK_PATH = '/v1/track'
...
+    constructor(message: string) {
+        super(message)
+        this.name = 'ResponseError'
+    }
+
+    status: number
+}
+
+const exponentialBackoff: RequestInitWithRetry['retryDelay'] = (
+    attempt, error, response
```
jonathannorris, line 27: These two params don't seem to be used: `error, response`

sdk/nodejs/src/request.ts
```diff
@@ -43,87 +14,119 @@ const TRACK_PATH = '/v1/track'
...
+
+type retryOnRequestErrorFunc = (retries: number) => RequestInitWithRetry['retryOn']
+
+const retryOnRequestError: retryOnRequestErrorFunc = (retries) => {
+    return (attempt, error, response) => {
+        if (attempt >= retries) {
+            return false
+        }
+
+        if (response && response?.status < 500) {
```
jonathannorris, line 40: Should we handle for 404s here?

sdk/nodejs/src/request.ts
```diff
@@ -43,87 +14,119 @@ const TRACK_PATH = '/v1/track'
...
+
+        return true
+    }
+}
+
+const handleResponse = async (res: Response) => {
+    if (!res.ok) {
+        let error
+        try {
+            error = new ResponseError((await res.clone().json()).message || 'Something went wrong')
```
jonathannorris, line 52: why the need for res.clone()

sdk/nodejs/src/request.ts
ajwootto: handle them how? this is basically just making sure we only retry retryable errors (5xx)

sdk/nodejs/src/request.ts
```diff
@@ -43,87 +14,119 @@ const TRACK_PATH = '/v1/track'
...
-        data: { batch: eventsBatch }
-    }, envKey, false)
-    if (res.status !== 201) {
-        logger.error(`Error posting events, status: ${res.status}, body: ${res.data?.message}`)
-    } else {
-        logger.debug(`Posted Events, status: ${res.status}, body: ${res.data?.message}`)
-    }
-
-    return res
+    return await post(url, {
```
jonathannorris, line 73: Do we need to set `application/json` headers?

ajwootto: it is being set

ajwootto: for POST anyway..

tsconfig.base.json
```diff
@@ -10,7 +10,7 @@
     "experimentalDecorators": true,
     "noImplicitAny": true,
     "importHelpers": false,
-    "target": "es5",
+    "target": "es2015",
```
jonathannorris, line 13: why this target when `es2019` is targeted in the lib below.

jonathannorris: Can we keep these targets consistent? also can we make it newer then this?

jonathannorris (changes requested):

tsconfig.base.json
ajwootto: they're different because the `lib` setting defines what type definitions are loaded, while this defines how the actual transpilation to plain JS is done. We can potentially target a newer version but we're still trying to support fairly old NodeJS versions. According to this table we could probably get away with ES2019?
https://node.green/

sdk/nodejs/tsconfig.spec.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "./tsconfig.json",
   "compilerOptions": {
+    "target": "es2017",
```
jonathannorris, line 4: should this be es2019 as well?

jonathannorris (approved):

PR #296, opened 2022-10-11
chore: added delay before closing the connection if the page focus is restored

jsalaber:
# Summary

Store the id of the `setTimeout` that occurs when page visibility changes to anything but visible. If the page becomes visible before `setTimeout` timeout occurs it will check if it needs to reconnect and then it will clear the `setTimeout` timer.

### DIFF
--- a/sdk/js/src/Client.ts
+++ b/sdk/js/src/Client.ts
@@ -38,6 +38,7 @@ export class DVCClient implements Client {
     eventEmitter: EventEmitter
     private streamingConnection?: StreamingConnection
     private pageVisibilityHandler?: () => void
+    private inactivityHandlerId?: number
 
     constructor(environmentKey: string, user: DVCUser, options: DVCOptions = {}) {
         this.user = new DVCPopulatedUser(user, options)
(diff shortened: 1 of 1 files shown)


sdk/js/src/Client.ts
```diff
@@ -300,22 +301,32 @@ export class DVCClient implements Client {
...
                     this.logger.debug('Page became visible, refetching config')
                     this.refetchConfig().catch((e) => {
                         this.logger.error('Failed to refetch config', e)
                     })
                     this.streamingConnection?.reopen()
-                } else {
+                }
+                window?.clearTimeout(this.inactivityHandlerId)
+            } else {
+                this.inactivityHandlerId = window?.setTimeout(() => {
```
ajwootto, line 322: just to be safe you may want to clear the timeout here as well if it exists, before reassigning to a new id. There may be situations where some combination of events causes this to leave a leftover timer

ajwootto: I wonder if it's also worth checking specifically for the visibility state being "hidden" rather than using an "else" here.

jsalaber: according to this there's only two states, `hidden` and `visible`: https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilityState

ajwootto (approved):

PR #297, opened 2022-10-11
add poweredByAlignment to project opt-in settings type

chantalchung:


lib/shared/types/src/types/config/models/project.ts
```diff
@@ -25,6 +25,7 @@ export class Project<IdType = string> {
                 primary: string
                 secondary: string
             }
+            poweredByPosition?: 'center' | 'left' | 'right'
```
laurawarr, line 28: kind of a nitpick, but could we change this to `alignment`? "position" makes me think top/bottom/left/right, but this seems like it's actually controlling the alignment along the bottom?

chantalchung: yup, will update

laurawarr (approved):

PR #299, opened 2022-10-11
chore(build): adds build:cdn step

nikolasleblanc:


package.json
```diff
@@ -19,6 +19,7 @@
   },
   "dependencies": {
     "@nestjs/class-validator": "^0.13.4",
+    "@nrwl/webpack": "^14.8.3",
```
ajwootto, line 22: this is a dev dependency

ajwootto: also make sure it's the same version as all the other @nrwl packages

ajwootto (approved):

PR #301, opened 2022-10-12
add instrumentation

andrewtraviss:


jonathannorris (changes requested):

jonathannorris (approved):

sdk/nodejs/src/eventQueue.ts
```diff
@@ -38,15 +38,17 @@ export type EventQueueOptions = {
...
+    private readonly reporter?: DVCReporter
     private readonly environmentKey: string
     eventFlushIntervalMS: number
     flushEventQueueSize: number
     maxEventQueueSize: number
     private flushInterval: NodeJS.Timer
     private flushInProgress = false
 
-    constructor(logger: DVCLogger, environmentKey: string, options: EventQueueOptions = {}) {
+    constructor(logger: DVCLogger, environmentKey: string, options: EventQueueOptions = {}, reporter?: DVCReporter) {
```
ajwootto, line 49: couldn't this just be added as an options field rather than another constructor arg

lib/shared/types/src/reporter.ts
```diff
@@ -0,0 +1,16 @@
+
+export type FlushResults = {
+    successes: number,
+    failures: number,
+    retries: number,
+}
+
+export type Tags = Record<string, string>
+
+export interface DVCReporter {
```
jonathannorris, line 10: Should we change this Reporter interface to be more generic so we don't need to add a new function / type here for every new metric?

sdk/nodejs/src/eventQueue.ts
jonathannorris: Also where is this actually set? I don't see where `DVCOptions.reporter` is passed into this reporter.

jonathannorris: Also DVCOptions is already being passed in here, so I think we should just grab it from there: 
image

lib/shared/types/src/reporter.ts
andrewtraviss: I tried to keep the amount of data formatting built into the API to a minimum, but yes most of these should be covered by a `reportMetricValue` method with a key.

I think it makes sense to keep `reportFlushResults` as a bespoke call, because the fields on that are deeply related and the api shouldn't predict how they might be used together.

sdk/nodejs/src/eventQueue.ts
andrewtraviss: Ah, missed this link in the chain. Added.

sdk/nodejs/__tests__/eventQueue.test.ts
```diff
@@ -5,13 +5,14 @@ import { AxiosResponse } from 'axios'
 import { EventQueue, EventQueueOptions } from '../src/eventQueue'
 import { EventTypes } from '../src/eventQueue'
 import { publishEvents } from '../src/request'
-import { BucketedUserConfig, DVCLogger, PublicProject } from '@devcycle/types'
+import { BucketedUserConfig, DVCLogger, DVCReporter, FlushResults, PublicProject } from '@devcycle/types'
```
jonathannorris, line 8: based on the warning

```suggestion
import { BucketedUserConfig, DVCLogger, DVCReporter, PublicProject } from '@devcycle/types'
```

sdk/nodejs/__tests__/eventQueue.test.ts
```diff
@@ -5,13 +5,14 @@ import { AxiosResponse } from 'axios'
...
 import { mocked } from 'jest-mock'
 import { dvcDefaultLogger } from '../src/utils/logger'
 import { cleanupBucketingLib, getBucketingLib, importBucketingLib } from '../src/bucketing'
 import { setPlatformDataJSON } from './utils/setPlatformData'
 
 import testData from '@devcycle/bucketing-test-data/json-data/testData.json'
+import { report } from 'process'
```
jonathannorris, line 15: based on the warning
```suggestion
```

sdk/nodejs/src/client.ts
```diff
@@ -51,7 +51,8 @@ export class DVCClient {
                 this.eventQueue = new EventQueue(
                     this.logger,
                     environmentKey,
-                    options
+                    options,
+                    options?.reporter
```
jonathannorris, line 55: If we are already passing in `options` why can't we just read the `reporter` from there?

andrewtraviss: Yeah I was just looking at this. The same applies for the logger.

andrewtraviss: Passing both through options now

PR #302, opened 2022-10-13
jssdk: refetch config when iframe emits message

chantalchung:
- listen for `DVC.optIn.saved` event message

### DIFF
--- a/sdk/js/src/Client.ts
+++ b/sdk/js/src/Client.ts
@@ -75,6 +75,14 @@ export class DVCClient implements Client {
                 this.eventEmitter.emitError(err)
                 return this
             })
+            
+        if (!options?.reactNative && typeof window !== 'undefined') {
+            window.addEventListener('message', (event) => {
+                if (event.data === 'DVC.optIn.saved') {
+                    this.refetchConfig()
+                }
+            })
+        }
     }
 
     onClientInitialized(): Promise<DVCClient>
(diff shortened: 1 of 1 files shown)


sdk/js/src/Client.ts
```diff
@@ -75,6 +75,14 @@ export class DVCClient implements Client {
                 this.eventEmitter.emitError(err)
                 return this
             })
+            
+        if (!options?.reactNative && typeof window !== 'undefined') {
+            window.addEventListener('message', (event) => {
+                if (event.data === 'DVC.optIn.saved') {
```
laurawarr, line 81: should we update this and the opt-in widget to use the object structure?
```suggestion
                const message = event.data
                if (message?.type === 'DVC.optIn.saved') {
```

chantalchung: yup, will make a PR to update the message it in the widget

laurawarr (approved):

sdk/js/src/Client.ts
```diff
@@ -75,6 +75,15 @@ export class DVCClient implements Client {
                 this.eventEmitter.emitError(err)
                 return this
             })
+            
+        if (!options?.reactNative && typeof window !== 'undefined') {
+            window.addEventListener('message', (event) => {
```
ajwootto, line 80: Can you also unregister this listener in the `close` method?

chantalchung: sorry, I'll make a PR to add this!

PR #303, opened 2022-10-13
DVC-4812 Add axios-retry to NodeJS SDK, use exponentialDelay and timeout

jonathannorris:


sdk/js/src/Request.ts
```diff
@@ -6,13 +6,12 @@ import axiosRetry from 'axios-retry'
 import { BucketedUserConfig, DVCLogger } from '@devcycle/types'
 
 const axiosClient = axios.create({
+    timeout: 60 * 1000,
```
ajwootto, line 9: isn't 60 seconds way too long of a timeout?

sdk/nodejs/src/request.ts
```diff
@@ -1,11 +1,21 @@
...
 
 const axiosClient = axios.create({
+    timeout: 60 * 1000,
     validateStatus: (status: number) => status < 400 && status >= 200,
 })
+axiosRetry(axiosClient, {
+    // will do exponential retry until axios.timeout
+    retryDelay: axiosRetry.exponentialDelay,
+    retryCondition: (error) => {
+        return (error?.response?.status || 0) >= 500
```
ajwootto, line 15: this should also retry if there's no response on the error, otherwise this won't actually fix the problem (which is a request timeout)

sdk/nodejs/src/request.ts
```diff
@@ -1,11 +1,21 @@
 import axios, { AxiosRequestConfig, AxiosRequestHeaders, AxiosResponse } from 'axios'
+import axiosRetry from 'axios-retry'
 import { SDKEventBatchRequestBody, DVCLogger } from '@devcycle/types'
 import { DVCPopulatedUser } from './models/populatedUser'
 import { DVCEvent, DVCOptions } from './types'
 
 const axiosClient = axios.create({
+    timeout: 60 * 1000,
```
ajwootto, line 8: this should be like 2 seconds IMO

sdk/js/src/Request.ts
jonathannorris: I set it to 60 because I think that's the timeout of our ALBs. We need some timeout set, so that the exponential backoff stops. By default, there is no timeout in Axios.

jonathannorris: We could set a limited number of retries on the axios-retry logic.

sdk/js/src/Request.ts
```diff
@@ -6,13 +6,12 @@ import axiosRetry from 'axios-retry'
...
     validateStatus: (status: number) => status < 400 && status >= 200,
```
jonathannorris, line 10: The one problem for this with `status < 400` is that 404s for invalid configs would get retried...

sdk/nodejs/src/request.ts
jonathannorris: I guess I should pass that timeout through to this.

sdk/js/src/Request.ts
ajwootto: this timeout is by individual request isn't it? I imagine you'd want the first request to timeout much faster than 60 seconds

sdk/js/src/Request.ts
ajwootto: isn't that what retryCondition is for? also pretty sure that case is a 401 Unauthorized

sdk/js/src/Request.ts
jonathannorris: I don't think so, there is a `shouldResetTimeout` option that defaults to `false`: https://github.com/softonic/axios-retry#options

sdk/js/src/Request.ts
```diff
@@ -6,13 +6,12 @@ import axiosRetry from 'axios-retry'
...
 })
 axiosRetry(axiosClient, {
-    retries: 5,
-    retryDelay: (retryCount) => {
-        return Math.pow(2, retryCount) * 1000
-    },
+    // will do exponential retry until axios.timeout
+    retryDelay: axiosRetry.exponentialDelay,
     retryCondition: (error) => {
         return (error?.response?.status || 0) >= 500
```
ajwootto, line 16: also retry if there's no response here

sdk/nodejs/src/request.ts
jonathannorris: yea its already passed through, changing the default to 10 sec? 
image

sdk/js/src/Request.ts
```diff
@@ -6,15 +6,14 @@ import axiosRetry from 'axios-retry'
...
 axiosRetry(axiosClient, {
-    retries: 5,
-    retryDelay: (retryCount) => {
-        return Math.pow(2, retryCount) * 1000
-    },
+    // will do exponential retry until axios.timeout
+    retryDelay: axiosRetry.exponentialDelay,
     retryCondition: (error) => {
-        return (error?.response?.status || 0) >= 500
+        return !error.response || (error?.response?.status || 0) >= 500
```
ajwootto, line 16: no point on null checking error in the second statement if you aren't in the first one

lib/shared/types/src/types/apis/sdk/clientSDKAPI.ts
```diff
@@ -323,7 +323,7 @@ export class DVCOptInUser {
     @IsString()
     @IsNotBlank()
     @IsNotEmpty()
-    user_id: string
+        user_id: string
```
laurawarr, line 326: why the extra indentation here?

jonathannorris: it was a linting change.

sdk/nodejs/src/request.ts
```diff
@@ -1,11 +1,23 @@
...
 import { SDKEventBatchRequestBody, DVCLogger } from '@devcycle/types'
 import { DVCPopulatedUser } from './models/populatedUser'
 import { DVCEvent, DVCOptions } from './types'
 
 const axiosClient = axios.create({
+    timeout: 10 * 1000,
     validateStatus: (status: number) => status < 400 && status >= 200,
 })
+axiosRetry(axiosClient, {
+    retries: 5,
```
ajwootto, line 12: the problem now is that the node SDK is still refetching the config every 10 seconds from the environmentConfigManager, so if theres a lot of request failures happening this will now pile up an increasing set of retrying requests

jonathannorris: so one option would be to make the config fetch timeout be 1 sec, and I think 5 retries would still then take less than 5 sec to complete. But the problem is that this is kinda a global config that applies to all requests...

jonathannorris: - Setup multiple axios clients for retryable requests, and non-retryable requests (events)
- Only log error if config fetch errors on first fetch.

jsalaber (approved):

sdk/nodejs/src/request.ts
ajwootto: won't this still be a problem with the latest changes since you'll still be retrying a timed out request up to 5 times while still proceeding with the 10 second config fetching loop

jonathannorris: added a new axios client for config requests that will do a single retry.

ajwootto (approved):

PR #304, opened 2022-10-13
remove event listener on message in close method

chantalchung:


sdk/js/src/Client.ts
```diff
@@ -246,6 +248,10 @@ export class DVCClient implements Client {
             document.removeEventListener('visibilitychange', this.pageVisibilityHandler)
         }
 
+        if (typeof window !== 'undefined' && this.windowMessageHandler) {
```
ajwootto, line 251: does this need the react native check as well? You could also just only assign the windowMessageHandler property if its being registered above, then just check if it exists here

chantalchung: yeah, I'll update so I only assign the `windodwMessageHandler` when we add the event handler

ajwootto (approved):

PR #309, opened 2022-10-25
DVC-4916 update fetchConfig interval to handle errors better

jonathannorris:
- Only throw an error on the initial config error.
- Log error for subsequent intervals if the initial config failed to load.


lib/shared/bucketing-assembly-script/index.js
```diff
@@ -15,6 +15,12 @@ exports.instantiate = async () => {
         }
     })()
 
+    const memory = new WebAssembly.Memory({
```
jonathannorris, line 18: cleaning up

sdk/nodejs/src/environmentConfigManager.ts
```diff
@@ -40,8 +40,13 @@ export class EnvironmentConfigManager {
 
         this.fetchConfigPromise = this._fetchConfig().then(() => {
             this.logger.debug('DevCycle initial config loaded')
+        }).finally(() => {
+            try {
+                this.intervalTimeout = setInterval(() => this._fetchConfig(), this.pollingIntervalMS)
```
ajwootto, line 45: you should try catch inside the setInterval handler function not at this level

ajwootto (approved):

PR #310, opened 2022-10-26
TAP-8702 report WASM memory usage

jonathannorris:
- report on WASM memory usage every 30 sec
- allow WebAssembly.MemoryDescriptor to be set as start option
- Change default initial WASM memory to 1,000 pages = 64 mb, max 10,000 pages = 640 mb

### DIFF
--- a/sdk/nodejs/src/bucketing.ts
+++ b/sdk/nodejs/src/bucketing.ts
@@ -1,8 +1,37 @@
 import { instantiate, Exports } from '@devcycle/bucketing-assembly-script'
+import { DVCLogger, DVCReporter } from '@devcycle/types'
+import { DVCOptions } from './types'
 
 let Bucketing: Exports | null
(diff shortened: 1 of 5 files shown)


sdk/nodejs/src/bucketing.ts
```diff
@@ -1,8 +1,38 @@
...
+import { DVCLogger, DVCReporter } from '@devcycle/types'
+import { DVCOptions } from './types'
 
 let Bucketing: Exports | null
-export const importBucketingLib = async (): Promise<void> => {
-    Bucketing = await instantiate()
+
+export const importBucketingLib = async (
+    { logger, options }:
+    { logger?: DVCLogger, options?: DVCOptions }
```
andrewtraviss, line 9: Since both parameters are optional, this can be defaulted to `{}` to simplify a few things.

andrewtraviss (approved):

PR #312, opened 2022-11-01
DVC-4916 - imports specific lodash modules

nikolasleblanc:
Reduces lodash's impact on the bundle size of the node library and the cdn file by importing only the lodash methods required by the code

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,11 +1,12 @@
 {
   "root": true,
   "ignorePatterns": ["**/*"],
-  "plugins": ["@nrwl/nx"],
+  "plugins": ["@nrwl/nx", "lodash"],
   "overrides": [
     {
       "files": ["*.ts", "*.tsx", "*.js", "*.jsx"],
       "rules": {
+        "lodash/import-scope": [2, "method"],
         "@nrwl/nx/enforce-module-boundaries": [
           "error",
           {
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -6,7 +6,7 @@
     {
       "files": ["*.ts", "*.tsx", "*.js", "*.jsx"],
       "rules": {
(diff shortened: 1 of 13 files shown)


lib/shared/bucketing/src/bucketing.ts
```diff
@@ -1,5 +1,8 @@
 'use strict'
-import { orderBy, pick, last, first } from 'lodash'
```
ajwootto, line 2: can you add this lint rule to prevent this in the future?
https://github.com/wix/eslint-plugin-lodash/blob/master/docs/rules/import-scope.md

nikolasleblanc: Oh good call

nikolasleblanc: Added

.eslintrc.json
```diff
@@ -1,11 +1,12 @@
 {
   "root": true,
   "ignorePatterns": ["**/*"],
-  "plugins": ["@nrwl/nx"],
+  "plugins": ["@nrwl/nx", "lodash"],
   "overrides": [
     {
       "files": ["*.ts", "*.tsx", "*.js", "*.jsx"],
       "rules": {
+        "lodash/import-scope": [2, "method"],
```
ajwootto, line 9: should this be configured to error instead of warn?

nikolasleblanc: Weird, it was showing as error.

ajwootto: thats fine then, i was just basing this on their docs

nikolasleblanc: Oh it's the 2, that means error.

nikolasleblanc: I updated it to "error" so its more clear

ajwootto (approved):

PR #313, opened 2022-11-02
feat: add type-safe generic for variable class to JS SDK

ajwootto:
- add a generic that produces the correct type for Variable objects based on the default value
- return default value in case where the config type does not match the code's type
- add tests for both


ajwootto:
* main
  * **PR #313** <a href="https://app.graphite.dev/github/pr/DevCycleHQ/js-sdks/313" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/DevCycleHQ/js-sdks/313?utm_source=stack-comment).

jsalaber (approved):

lib/shared/types/src/types/config/models/variable.ts
```diff
@@ -41,3 +41,10 @@ export enum VariableType {
  */
 export type VariableValue = string | boolean | number | JSON
 type JSON = { [key: string]: string | boolean | number }
+
+// alias to resolve a generic type constrained by `VariableValue` back into its original type
+export type VariableTypeAlias<T> = T extends boolean
```
jonathannorris, line 46: any way to format this with brackets so it makes a bit more sense?

PR #316, opened 2022-11-03
fix: specifying libs dir breaks next app builds in Nx

ajwootto:
Working around this Nx bug
https://github.com/nrwl/nx/issues/9317


jsalaber (approved):
damn all that struggling and for it to be only one line 😭

ajwootto:
> damn all that struggling and for it to be only one line 😭
welcome to nx hehe

PR #318, opened 2022-11-07
DVC-4962

nikolasleblanc:
Reduces bundle size by ensuring the types package is treeshakable.

Also includes:

- local verdaccio install for local testing outside the monorepo and docs

### DIFF
--- a/lib/shared/types/project.json
+++ b/lib/shared/types/project.json
@@ -18,18 +18,25 @@
       }
     },
     "build": {
-      "executor": "@nrwl/js:tsc",
+      "executor": "@nrwl/web:rollup",
       "outputs": [
         "{options.outputPath}"
       ],
       "options": {
+        "format": ["esm", "cjs"],
+        "project": "lib/shared/types/package.json",
         "outputPath": "dist/lib/shared/types",
-        "main": "lib/shared/types/src/index.ts",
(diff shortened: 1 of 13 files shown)


lib/shared/types/tsconfig.json
```diff
@@ -1,7 +1,9 @@
 {
   "extends": "../../../tsconfig.base.json",
   "compilerOptions": {
-    "module": "CommonJS",
+    "target": "es2015",
```
ajwootto, line 4: can we target a more recent version for this?

lib/shared/types/project.json
```diff
@@ -18,18 +18,25 @@
...
         "assets": [
-          "lib/shared/types/*.md"
-        ]
+          {
+            "glob": "*.md",
+            "input": "lib/shared/types",
+            "output": "."
+          }
+        ],
+        "skipTypeField": true
```
ajwootto, line 39: does it not work with that field?

lib/shared/types/tsconfig.json
nikolasleblanc: Think so, I'll give it a shot.

lib/shared/types/project.json
ajwootto: weird..

ajwootto (approved):

PR #323, opened 2022-11-15
DVC 5112 store and read anonymousId

L-Joli:
- add store and read anonymous id logic to `Client`

### DIFF
--- a/sdk/js/__tests__/Client.spec.js
+++ b/sdk/js/__tests__/Client.spec.js
@@ -135,16 +135,23 @@ describe('DVCClient tests', () => {
         expect(client.config).toBeUndefined()
     })
 
-    it('should have same value for anonymous user id from local storage and user_id if anonId exists and isAnonymous is true', async () => {
+    it('should have same value for anonymous user id from local storage and user_id if anonId exists and isAnonymous is true', () => {
         const client = new DVCClient('test_env_key', { isAnonymous: true })
-        const anonymousUserId = await client.store.load(StoreKey.AnonUser)
+        const anonymousUserId = client.store.load(StoreKey.AnonUser)
         expect(JSON.parse(anonymousUserId)).toEqual(client.user.user_id)
     })
 
-    it('should not save anonymous user id in local storage if isAnonymous is false', async () => {
+    it('should not save anonymous user id in local storage if isAnonymous is false', () => {
         const client = new DVCClient('test_env_key', { user_id: 'user1', isAnonymous: false })
-        const anonymousUserId = await client.store.load(StoreKey.AnonUser)
+        const anonymousUserId = client.store.load(StoreKey.AnonUser)
         expect(anonymousUserId).toBeNull()
+    })
+
+    it('should get the anonymous user id from local storage if it exists', () => {
+        Storage.prototype.getItem = jest.fn(() => 'test_anon_user_id');
+        const client = new DVCClient('test_env_key', { isAnonymous: true })
+        expect(client.user.user_id).toEqual("test_anon_user_id")
+        global.Storage.prototype.getItem.mockReset()
     })
 
     describe('onClientInitialized', () => {
--- a/sdk/js/__tests__/Client.spec.js
+++ b/sdk/js/__tests__/Client.spec.js
@@ -3,6 +3,7 @@ import { getConfigJson, publishEvents, saveEntity } from '../src/Request'
 import { mocked } from 'jest-mock'
 import { DVCVariable } from '../src/Variable'
 import { DVCPopulatedUser } from '../src/User'
+import { StoreKey } from '../src/Store'
 
 jest.mock('../src/Request')
 jest.mock('../src/StreamingConnection')
@@ -132,6 +133,18 @@ describe('DVCClient tests', () => {
         expect(getConfigJson_mock).toBeCalled()
         await client.onInitialized
         expect(client.config).toBeUndefined()
+    })
+
+    it('should have same value for anonymous user id from local storage and user_id if anonId exists and isAnonymous is true', async () => {
+        const client = new DVCClient('test_env_key', { isAnonymous: true })
+        const anonymousUserId = await client.store.load(StoreKey.AnonUser)
+        expect(JSON.parse(anonymousUserId)).toEqual(client.user.user_id)
+    })
+
+    it('should not save anonymous user id in local storage if isAnonymous is false', async () => {
+        const client = new DVCClient('test_env_key', { user_id: 'user1', isAnonymous: false })
+        const anonymousUserId = await client.store.load(StoreKey.AnonUser)
+        expect(anonymousUserId).toBeNull()
     })
 
     describe('onClientInitialized', () => {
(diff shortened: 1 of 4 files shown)


sdk/js/src/Client.ts
```diff
@@ -46,7 +46,20 @@ export class DVCClient implements Client {
     private windowMessageHandler?: (event: MessageEvent) => void
 
     constructor(environmentKey: string, user: DVCUser, options: DVCOptions = {}) {
+        this.logger = options.logger || dvcDefaultLogger({ level: options.logLevel })
+        this.store = new Store(typeof window !== 'undefined' ? window.localStorage : stubbedLocalStorage, this.logger)
+        
+        const storedAnonymousId = this.store.store?.getItem(StoreKey.AnonUser)
```
rodeora, line 52: it would be better to use `this.store.load` function instead

sdk/js/__tests__/Client.spec.js
```diff
@@ -134,6 +135,24 @@ describe('DVCClient tests', () => {
         expect(client.config).toBeUndefined()
     })
 
+    it('should have same value for anonymous user id from local storage and user_id if anonId exists and isAnonymous is true', async () => {
+        const client = new DVCClient('test_env_key', { isAnonymous: true })
+        const anonymousUserId = client.store.store.getItem(StoreKey.AnonUser)
```
rodeora, line 140: use client.store.load instead

sdk/js/__tests__/Client.spec.js
```diff
@@ -134,6 +135,24 @@ describe('DVCClient tests', () => {
...
+
+    it('should save anonymous user id in local storage if isAnonymous is true', async () => {
+        const client = new DVCClient('test_env_key', { isAnonymous: true })
+        const anonymousUserId = client.store.store.getItem(StoreKey.AnonUser)
+        expect(anonymousUserId).toBeTruthy()
+    })
+
+    it('should not save anonymous user id in local storage if isAnonymous is false', async () => {
+        const client = new DVCClient('test_env_key', { user_id: 'user1', isAnonymous: false })
+        const anonymousUserId = client.store.store.getItem(StoreKey.AnonUser)
```
rodeora, line 152: same here

sdk/js/__tests__/Client.spec.js
```diff
@@ -134,6 +135,24 @@ describe('DVCClient tests', () => {
...
+    it('should have same value for anonymous user id from local storage and user_id if anonId exists and isAnonymous is true', async () => {
+        const client = new DVCClient('test_env_key', { isAnonymous: true })
+        const anonymousUserId = client.store.store.getItem(StoreKey.AnonUser)
+        expect(JSON.parse(anonymousUserId)).toEqual(client.user.user_id)
+    })
+
+    it('should save anonymous user id in local storage if isAnonymous is true', async () => {
+        const client = new DVCClient('test_env_key', { isAnonymous: true })
+        const anonymousUserId = client.store.store.getItem(StoreKey.AnonUser)
+        expect(anonymousUserId).toBeTruthy()
```
rodeora, line 147: this is testing the same thing as the previous test, you can delete this

rodeora (changes requested):

sdk/js/src/Client.ts
L-Joli: ummm...but it returns a Promise, do we really need a Promise for this case?

L-Joli: I can clear all the promises in store in other pr

sdk/js/src/Store.ts
```diff
@@ -22,10 +22,8 @@ export class Store {
         }, storeKey)
     }
 
-    load(storeKey: string): Promise<string | null | undefined> {
-        return new Promise((resolve) => {
-            resolve(this.store?.getItem(storeKey))
-        })
+    load(storeKey: string): string | null | undefined{
```
rodeora, line 25: can you un-promisify everything in here?

rodeora (changes requested):

sdk/js/src/Store.ts
```diff
@@ -16,53 +16,38 @@ export class Store {
...
+            resolve(this.save(StoreKey.Config, data))
+        })
     }
 
-    loadConfig(): Promise<string | null | undefined> {
+    loadConfig(): string | null | undefined {
         return this.load(StoreKey.Config)
     }
 
     saveUser(user: DVCPopulatedUser): Promise<void> {
```
rodeora, line 41: this shouldn't be a promise either

sdk/js/src/Store.ts
```diff
@@ -16,53 +16,38 @@ export class Store {
...
-            })
-        } catch (e) {
-            this.logger?.error(e as string)
-            return Promise.resolve()
-        }
+    remove(storeKey: string): void {
+        this.store?.removeItem(storeKey)
     }
 
     saveConfig(data: BucketedUserConfig): Promise<void> {
```
rodeora, line 31: remove the promise here

rodeora (changes requested):

sdk/js/src/Store.ts
L-Joli: but in `Client.ts`, seems like we want to log the success message. If it's not a promise, we don't know if it's succeeded. Do we want to keep the logger and promise or just remove both? same for the `saveConfig` case
```
this.store.saveUser(user).then(() => {
       this.logger.info('Successfully saved user to local storage')
})
```

rodeora: you can just log it here instead of the client

sdk/js/src/Store.ts
```diff
@@ -16,53 +16,33 @@ export class Store {
...
-
-    loadConfig(): Promise<string | null | undefined> {
+    loadConfig(): string | null | undefined {
         return this.load(StoreKey.Config)
     }
 
-    saveUser(user: DVCPopulatedUser): Promise<void> {
-        if (!user) {
-            return Promise.reject('No user to save')
-        }
```
rodeora, line 57: still want to keep this, throw an error

rodeora (changes requested):

sdk/js/__tests__/Client.spec.js
```diff
@@ -134,6 +135,18 @@ describe('DVCClient tests', () => {
...
+        const anonymousUserId = client.store.load(StoreKey.AnonUser)
+        expect(JSON.parse(anonymousUserId)).toEqual(client.user.user_id)
+    })
+
+    it('should not save anonymous user id in local storage if isAnonymous is false', async () => {
+        const client = new DVCClient('test_env_key', { user_id: 'user1', isAnonymous: false })
+        const anonymousUserId = client.store.load(StoreKey.AnonUser)
+        expect(anonymousUserId).toBeNull()
+    })
+
```
rodeora, line 149: we need a test to check an anon_user gets the same id as the one in local storage

rodeora (changes requested):

sdk/js/__tests__/Client.spec.js
L-Joli: umm... this test `should have same value for anonymous user id from local storage and user_id if anonId exists and isAnonymous is true` includes the checking "anonymous user id =  the one in local storage". Is there something else I need to add into this?

rodeora: you need a new test, the existing one is checking that the anon user's id was saved to local storage. The new test needs to check that a new anon user's id gets set to the same value as the one saved in local storage

rodeora (approved):

PR #324, opened 2022-11-15
Dvc 5114 js sdk remove id from localstorage on reset

L-Joli:
- remove anonymous user id from local storage on reset
- branch out from and merge after https://github.com/DevCycleHQ/js-sdks/pull/323

### DIFF
--- a/sdk/js/__tests__/Client.spec.js
+++ b/sdk/js/__tests__/Client.spec.js
@@ -560,6 +560,17 @@ describe('DVCClient tests', () => {
 
             expect(publishEvents).toBeCalled()
         })
+
+        it('should remove anonymous user id from local storage', async () => {
+            const client = new DVCClient('test_env_key', { isAnonymous: true })
+            const oldAnonymousId = client.store.load(StoreKey.AnonUser)
+            expect(oldAnonymousId).toBeTruthy()
+
+            await client.resetUser()
+            const newAnonymousId = client.store.load(StoreKey.AnonUser)
+            expect(oldAnonymousId).not.toEqual(newAnonymousId)
+            expect(client.user.user_id).toEqual(JSON.parse(newAnonymousId))
+        })
     })
 
     describe('allFeatures', () => {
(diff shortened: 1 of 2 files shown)


sdk/js/src/Client.ts
```diff
@@ -208,8 +219,8 @@ export class DVCClient implements Client {
     resetUser(): Promise<DVCVariableSet>
     resetUser(callback: ErrorCallback<DVCVariableSet>): void
     resetUser(callback?: ErrorCallback<DVCVariableSet>): Promise<DVCVariableSet> | void {
+        this.store.store?.removeItem(StoreKey.AnonUser)
```
rodeora, line 222: use the store.clear function instead

sdk/js/src/Client.ts
```diff
@@ -208,8 +219,8 @@ export class DVCClient implements Client {
...
         const anonUser = new DVCPopulatedUser({ isAnonymous: true }, this.options)
-
```
rodeora, line 212: we also need to save this new anonUser's id to the store here

sdk/js/__tests__/Client.spec.js
```diff
@@ -494,6 +495,16 @@ describe('DVCClient tests', () => {
...
         })
+
+        it('should remove anonymous user id from local storage', async () => {
+            const client = new DVCClient('test_env_key', { isAnonymous: true })
+            let anonymousId = await client.store.load(StoreKey.AnonUser)
+            expect(anonymousId).toBeTruthy()
+            
+            await client.resetUser()
+            anonymousId = await client.store.load(StoreKey.AnonUser)
+            expect(anonymousId).toBeNull()
```
rodeora, line 506: we should be checking that the new anon user created by reset() exists here

rodeora (changes requested):

kaushalkapasi (approved):

rodeora (approved):

PR #325, opened 2022-11-15
fix: type definitions for example apps, write tests and change JSON type export

ajwootto:
- fix React type definition to work in example apps
- write tests for useVariableValue to test types
- rename `JSON` type to `DVCJSON` to prevent conflicts with global JSON


sdk/nodejs/src/client.ts
```diff
@@ -17,7 +17,6 @@ import * as packageJson from '../package.json'
 import { importBucketingLib, getBucketingLib, cleanupBucketingLib } from './bucketing'
 import { DVCLogger, getVariableTypeFromValue, VariableTypeAlias } from '@devcycle/types'
 import os from 'os'
-import { JSON } from '@devcycle/devcycle-js-sdk'
```
ajwootto, line 20: dunno how this snuck in here but this sets up a dependency on the JS SDK from the nodejs SDK..

nikolasleblanc (approved):

PR #326, opened 2022-11-16
jssdk remove anon_user from storage on identify

rodeora:


sdk/js/src/Client.ts
```diff
@@ -191,15 +191,23 @@ export class DVCClient implements Client {
             this.eventQueue.flushEvents()
 
             let updatedUser: DVCPopulatedUser
+            const bothAnonymous = this.user.isAnonymous && user.isAnonymous
+
             if (user.user_id === this.user.user_id) {
                 updatedUser = this.user.updateUser(user, this.options)
+            } else if (bothAnonymous) {
+                user.user_id = this.user.user_id
+                updatedUser = this.user.updateUser(user, this.options)
```
L-Joli, line 200: I don't think we need this part after making the change we've discussed in the channel, do we want to merge this and edit it later or remove it now?

rodeora: we'd still need this condition

L-Joli (approved):

PR #329, opened 2022-11-18
update DVCPopulatedUser to read anonymousUserId

L-Joli:
- update DVCPopulatedUser to read anonymousUserId

### DIFF
--- a/sdk/js/src/User.ts
+++ b/sdk/js/src/User.ts
@@ -26,13 +26,13 @@ export class DVCPopulatedUser implements DVCUser {
     readonly sdkType: 'client' | 'server'
     readonly sdkVersion: string
 
-    constructor(user: DVCUser, options: DVCOptions, staticData?: StaticData) {
+    constructor(user: DVCUser, options: DVCOptions, staticData?: StaticData, anonymousUserId?: string) {
         if (!user.user_id && !user.isAnonymous) {
             throw new Error('Must have a user_id, or have "isAnonymous" set on the user')
         }
 
-        this.user_id = !user.user_id ? uuidv4() : user.user_id
-        this.isAnonymous = user.user_id ? false : user.isAnonymous
+        this.user_id =  user.isAnonymous ? anonymousUserId || uuidv4() : user.user_id || uuidv4()
+        this.isAnonymous = user.isAnonymous || false
         this.email = user.email
         this.name = user.name
         this.language = user.language
--- a/sdk/js/src/User.ts
+++ b/sdk/js/src/User.ts
@@ -32,7 +32,7 @@ export class DVCPopulatedUser implements DVCUser {
         }
 
         this.user_id =  user.isAnonymous ? anonymousUserId || uuidv4() : user.user_id || uuidv4()
-        this.isAnonymous = user.isAnonymous || false
+        this.isAnonymous = user.user_id ? false : (user.isAnonymous ?? true)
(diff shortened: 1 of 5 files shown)


sdk/js/src/User.ts
```diff
@@ -26,13 +26,13 @@ export class DVCPopulatedUser implements DVCUser {
...
-    constructor(user: DVCUser, options: DVCOptions, staticData?: StaticData) {
+    constructor(user: DVCUser, options: DVCOptions, staticData?: StaticData, anonymousUserId?: string) {
         if (!user.user_id && !user.isAnonymous) {
             throw new Error('Must have a user_id, or have "isAnonymous" set on the user')
         }
 
-        this.user_id = !user.user_id ? uuidv4() : user.user_id
-        this.isAnonymous = user.user_id ? false : user.isAnonymous
+        this.user_id =  user.isAnonymous ? anonymousUserId || uuidv4() : user.user_id || uuidv4()
+        this.isAnonymous = user.isAnonymous || false
```
rodeora, line 35: blah || false always evaluates to blah, the extra false is redundant
```suggestion
        this.isAnonymous = user.isAnonymous
```

sdk/js/src/User.ts
```diff
@@ -26,13 +26,13 @@ export class DVCPopulatedUser implements DVCUser {
...
 
-    constructor(user: DVCUser, options: DVCOptions, staticData?: StaticData) {
+    constructor(user: DVCUser, options: DVCOptions, staticData?: StaticData, anonymousUserId?: string) {
         if (!user.user_id && !user.isAnonymous) {
             throw new Error('Must have a user_id, or have "isAnonymous" set on the user')
         }
 
-        this.user_id = !user.user_id ? uuidv4() : user.user_id
-        this.isAnonymous = user.user_id ? false : user.isAnonymous
+        this.user_id =  user.isAnonymous ? anonymousUserId || uuidv4() : user.user_id || uuidv4()
```
rodeora, line 34: If the user is not anonymous they always have an id
```suggestion
        this.user_id =  user.isAnonymous ? anonymousUserId || uuidv4() : user.user_id
```

rodeora (changes requested):

sdk/js/src/User.ts
L-Joli: I added this caz a test case `should make a new user if user id provided but no isAnonymous flag` was failed, should I update the test case then?

sdk/js/src/User.ts
L-Joli: yes, but there's a typescript error, `user_id` type is `undefined | string` so it will complain if not added a default value. I took a reference from android sdk and it uses `uuidv4()` to be the default value

sdk/js/src/User.ts
L-Joli: the test `expect(user.isAnonymous).toBe(false)`, but if I don't make a default value, `user.isAnonymous` will be undefined

L-Joli: ohh, i think because our tests are js file, so it didn't do type check... and actually adding the false value is also trying to follow the android sdk behaviour

sdk/js/src/User.ts
rodeora: oh nvm then, you can leave it as is then

rodeora (approved):

PR #331, opened 2022-11-22
Add caching for react native

kaushalkapasi:
- Adds caching functionality for React Native using `@react-native-async-storage/async-storage`
- Allows the creation of a `DVCPopulatedUser` without needing to specify a `user_id` OR `isAnonymous`
  - if no `user_id` is provided, we will automatically create an anonymous user

### DIFF
--- a/examples/react/react-native-app/ios/Podfile
+++ b/examples/react/react-native-app/ios/Podfile
@@ -2,7 +2,7 @@ require_relative '../node_modules/react-native/scripts/react_native_pods'
 require_relative '../node_modules/@react-native-community/cli-platform-ios/native_modules'
 require_relative '../node_modules/@nrwl/react-native/nx_post_install'
 
-platform :ios, '11.0'
+platform :ios, '13.0'
 
 target 'ReactNativeApp' do
   config = use_native_modules!
--- a/examples/react/react-native-app/src/app/App.tsx
+++ b/examples/react/react-native-app/src/app/App.tsx
@@ -30,6 +30,8 @@ const user = {
     isAnonymous: false
 }
 
+const anonUser = {}
+
 export const App = (): JSX.Element => {
     const [whatsNextYCoord, setWhatsNextYCoord] = useState<number>(0)
     const scrollViewRef = useRef<null | ScrollView>(null)
@@ -58,6 +60,15 @@ export const App = (): JSX.Element => {
                 console.log('Updated variables: ', variables)
             })
             .catch((err) => console.log({ err }))
+    }
+
+    const resetUser = () => {
+        console.log(`reset user`)
+        client && client.resetUser().catch(err => console.error(err))
+    }
+
+    const identifyAnonUser = () => {
+        client && client.identifyUser(anonUser)
     }
 
     if (!dvcReady) return <SafeAreaView><Text style={styles.textXL}>DVC is not ready!</Text></SafeAreaView>
@@ -94,6 +105,22 @@ export const App = (): JSX.Element => {
                         >
                             <Text style={[styles.textMd, styles.textCenter]}>
                   Identify new user
+                            </Text>
+                        </TouchableOpacity>
+                        <TouchableOpacity
+                            style={styles.whatsNextButton}
+                            onPress={() => resetUser()}
+                        >
+                            <Text style={[styles.textMd, styles.textCenter]}>
+                  Reset User
+                            </Text>
+                        </TouchableOpacity>
+                        <TouchableOpacity
+                            style={styles.whatsNextButton}
+                            onPress={() => identifyAnonUser()}
+                        >
+                            <Text style={[styles.textMd, styles.textCenter]}>
+                  Identify Anon User
                             </Text>
                         </TouchableOpacity>
                     </View>
--- a/examples/react/react-native-app/src/app/App.tsx
+++ b/examples/react/react-native-app/src/app/App.tsx
@@ -63,12 +63,11 @@ export const App = (): JSX.Element => {
     }
 
     const resetUser = () => {
-        console.log(`reset user`)
-        client && client.resetUser().catch(err => console.error(err))
+        client && client.resetUser().catch((err) => console.log({ err }))
     }
 
     const identifyAnonUser = () => {
-        client && client.identifyUser(anonUser)
+        client && client.identifyUser(anonUser).catch((err) => console.log({ err }))
     }
 
     if (!dvcReady) return <SafeAreaView><Text style={styles.textXL}>DVC is not ready!</Text></SafeAreaView>
--- a/sdk/js/__tests__/index.spec.js
+++ b/sdk/js/__tests__/index.spec.js
@@ -27,9 +27,10 @@ describe('initialize tests', () => {
         expect(() => DVCClient.initialize('YOUR_CLIENT_SIDE_ID')).toThrow(missingUserError)
     })
 
-    it('should throw an error if invalid user is passed in initialize', () => {
+    it('should NOT throw an error if invalid user is passed in initialize', () => {
         const badUser = { 'who': 'me' }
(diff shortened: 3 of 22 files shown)


examples/react/react-native-app/ios/Podfile
```diff
@@ -2,7 +2,7 @@ require_relative '../node_modules/react-native/scripts/react_native_pods'
 require_relative '../node_modules/@react-native-community/cli-platform-ios/native_modules'
 require_relative '../node_modules/@nrwl/react-native/nx_post_install'
 
-platform :ios, '11.0'
+platform :ios, '13.0'
```
jsalaber, line 5: what does this fix, does that mean we don't support ios under 13 for any react native app?

examples/react/react-native-app/src/app/App.tsx
```diff
@@ -60,6 +62,15 @@ export const App = (): JSX.Element => {
             .catch((err) => console.log({ err }))
     }
 
+    const resetUser = () => {
+        console.log(`reset user`)
```
jsalaber, line 66: console.log

jsalaber (changes requested):

examples/react/react-native-app/ios/Podfile
kaushalkapasi: Nope, this was just to run the app locally.. I didn't have an older version of `iOS` to test against. This should allow any developer running macOS Ventura (13.0) with XCode to run the example app without any issue

sdk/js/src/ReactNativeStore.ts
```diff
@@ -0,0 +1,58 @@
...
+        this.logger = logger
+    }
+
+    save(storeKey: string, data: unknown): void {
+        this.store.setItem(storeKey, JSON.stringify(data))
+    }
+
+    load(storeKey: string): any {
+        let storedValue
+        (async () => {
```
jsalaber, line 21: what's the reasoning behind this? is this going to make it an async function, why can't it just be an async function?

jsalaber (changes requested):

sdk/js/src/Client.ts
```diff
@@ -36,7 +37,7 @@ export class DVCClient implements Client {
     logger: DVCLogger
     config?: BucketedUserConfig
     user: DVCPopulatedUser
-    private store: Store
+    private store: Store | ReactNativeStore
```
jsalaber, line 40: could we make an abstract class so we don't have to do this?

jsalaber (changes requested):

sdk/js/src/Client.ts
kaushalkapasi: yeah! good callout :)

sdk/js/src/User.ts
```diff
@@ -23,16 +23,18 @@ export class DVCPopulatedUser implements DVCUser {
...
+            throw new Error('A User cannot be created with a user_id that is an empty string')
         }
-
-        this.user_id =  user.isAnonymous ? anonymousUserId || uuidv4() : user.user_id || uuidv4()
-        this.isAnonymous = user.user_id ? false : (user.isAnonymous ?? true)
+        
+        this.user_id =  (user.isAnonymous || !user.user_id) ? 
+            (anonymousUserId || uuidv4()) :
+            user.user_id
+        this.isAnonymous = (user.isAnonymous || !user.user_id) ? true : false
```
rodeora, line 37:
```suggestion
        this.isAnonymous = (user.isAnonymous || !user.user_id)
```

sdk/js/src/Client.ts
```diff
@@ -191,7 +198,7 @@ export class DVCClient implements Client {
             this.eventQueue.flushEvents()
 
             let updatedUser: DVCPopulatedUser
-            const bothAnonymous = this.user.isAnonymous && user.isAnonymous
+            const bothAnonymous = this.user.isAnonymous && (user.isAnonymous || true)
```
rodeora, line 201: (user.isAnonymous || true) always evaluates to true

sdk/js/src/User.ts
```diff
@@ -23,16 +23,14 @@ export class DVCPopulatedUser implements DVCUser {
...
-        if (!user.user_id && !user.isAnonymous) {
-            throw new Error('Must have a user_id, or have "isAnonymous" set on the user')
-        }
-
-        this.user_id =  user.isAnonymous ? anonymousUserId || uuidv4() : user.user_id || uuidv4()
-        this.isAnonymous = user.user_id ? false : (user.isAnonymous ?? true)
+        this.user_id =  (user.isAnonymous || !user.user_id) ? 
+            (anonymousUserId || uuidv4()) :
+            user.user_id
+        this.isAnonymous = user.user_id ? false : (user.isAnonymous || true)
```
L-Joli: actually if we use `user.isAnonymous || true`, when `isAnonymous` is false, it will fall to true. So should we just keep `user.user_id ? false : true` ?

sdk/js/src/Client.ts
kaushalkapasi: ah, this should be the same as the User constructor `isAnonymous`.. will update to `(user.isAnonymous || !user.user_id)`

sdk/js/__tests__/index.spec.js
```diff
@@ -27,9 +27,10 @@ describe('initialize tests', () => {
         expect(() => DVCClient.initialize('YOUR_CLIENT_SIDE_ID')).toThrow(missingUserError)
     })
 
-    it('should throw an error if invalid user is passed in initialize', () => {
+    it('should NOT throw an error if invalid user is passed in initialize', () => {
         const badUser = { 'who': 'me' }
```
L-Joli, line 31: why don't we want an error when the user looks like this?

sdk/js/src/Client.ts
```diff
@@ -51,18 +51,19 @@ export class DVCClient implements Client {
...
             this.store = new ReactNativeStore(this.logger)
         } else {
-            this.store = new Store(typeof window !== 'undefined' ? window.localStorage : stubbedLocalStorage, this.logger)
+            this.store = new Store(
+                typeof window !== 'undefined' ? window.localStorage : stubbedLocalStorage, this.logger
+            )
         }
 
-        if (!user.isAnonymous) {
+        if (user.user_id || !!!user.isAnonymous) {
```
L-Joli, line 59: `!!!undefined` === `true`, so `anonUserId` will be removed when user didn't provide both `user_id` and `isAnonymous`

kaushalkapasi: just updated this :) take a look again please! 🙏

sdk/js/src/User.ts
```diff
@@ -23,16 +23,18 @@ export class DVCPopulatedUser implements DVCUser {
...
+            throw new Error('A User cannot be created with a user_id that is an empty string')
         }
-
-        this.user_id =  user.isAnonymous ? anonymousUserId || uuidv4() : user.user_id || uuidv4()
-        this.isAnonymous = user.user_id ? false : (user.isAnonymous ?? true)
+        
+        this.user_id = (user.isAnonymous || !user.user_id) ? 
+            (anonymousUserId || uuidv4()) :
+            user.user_id
+        this.isAnonymous = (user.isAnonymous || !user.user_id)
```
rodeora, line 37: could simplify this, makes the conditional easier to read
```suggestion
        this.isAnonymous = (user.isAnonymous || !user.user_id)
        this.user_id = this.isAnonymous ? (anonymousUserId || uuidv4()) :user.user_id
```

kaushalkapasi: sadly, doesn't work due to the type inferencing... typescript will not allow this, as it can't tell that user.user_id would be defined... with the ternary in this line, it works because it checks of the existence in the `!user.user_id`

PR #335, opened 2022-11-29
DVC-0000: add sleuth deployment post

nikolasleblanc:


scripts/npm-safe-publish.sh
```diff
@@ -42,6 +44,12 @@ if [[ "$NPM_SHOW" != "$NPM_LS" ]]; then
   echo "Versions are not the same, (Remote = $NPM_SHOW; Local = $NPM_LS). Checking for publish eligibility."
 
   if [[ "$NPM_REGISTRY" = "https://registry.npmjs.org/" ]]; then
+    # make sure we're able to track this deployment
+    if [[ -z "$DEVCYCLE_PROD_SLEUTH_API_TOKEN" ]]; then
+      echo "Sleuth.io deployment tracking token not found. Aborting."
+      exit 1
```
JamieSinn, line 50: We shouldn't fail a deployment because of a failure to get the key this late - if it's truly required we should fail immediately

scripts/npm-safe-publish.sh
```diff
@@ -67,6 +75,11 @@ if [[ "$NPM_SHOW" != "$NPM_LS" ]]; then
     fi
 
     npm publish --otp=$OTP
+
+    curl -X POST \
```
JamieSinn, line 79: You don't check whether or not the npm publish step succeeded; or whether it actually published.
use `$?` to get the previous return code (in this case, npm)

JamieSinn (approved):

PR #339, opened 2022-12-01
Fix React Native Store import

kaushalkapasi:
- convert RN Store import to be a dynamic import only used when `options.reactNative` is set to `true`
- add missing dependencies for RN

### DIFF
--- a/sdk/js/src/Client.ts
+++ b/sdk/js/src/Client.ts
@@ -25,7 +25,6 @@ import { dvcDefaultLogger } from './logger'
 import { DVCLogger } from '@devcycle/types'
 import { StreamingConnection } from './StreamingConnection'
 import Store from './Store'
-import ReactNativeStore from './ReactNativeStore'
 
 export class DVCClient implements Client {
     private options: DVCOptions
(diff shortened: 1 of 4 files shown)


sdk/js/src/Client.ts
```diff
@@ -50,7 +49,10 @@ export class DVCClient implements Client {
     constructor(environmentKey: string, user: DVCUser, options: DVCOptions = {}) {
         this.logger = options.logger || dvcDefaultLogger({ level: options.logLevel })
         if (options?.reactNative) {
-            this.store = new ReactNativeStore(this.logger)
+            (async () => {
```
jonathannorris, line 52: if this is async, won't that cause all the `this.store` references below not to work?

jonathannorris: for example: 
```
const storedAnonymousId = this.store.loadAnonUserId()
```

sdk/react/package.json
```diff
@@ -17,6 +17,8 @@
   },
   "dependencies": {
     "@react-native-async-storage/async-storage": "^1.17.11",
-    "hoist-non-react-statics": "^3.3.2"
+    "hoist-non-react-statics": "^3.3.2",
+    "react-native-device-info": "^8.7.0",
```
L-Joli, line 21: for this package, we need to `import DeviceInfo from 'react-native-device-info'` and set `global.DeviceInfo = DeviceInfo` for RN according to our doc https://docs.devcycle.com/docs/sdk/client-side-sdks/react-native#installation. Can this also be done in our sdk? Or it will affect the react sdk so we shouldn't add it?

kaushalkapasi: yeah, we don't want it for React SDK.. so while we will install it, we won't set the `global.DeviceInfo` attribute

PR #340, opened 2022-12-02
[fix] remove RN caching and associated dependencies from JS & React SDKs

kaushalkapasi:
- remove RN caching to cut down on the package size, while we investigate & redesign the best way to implement this :)


laurawarr (approved):

sdk/js/src/Client.ts
```diff
@@ -49,13 +48,9 @@ export class DVCClient implements Client {
...
         this.logger = options.logger || dvcDefaultLogger({ level: options.logLevel })
-        if (options?.reactNative) {
-            this.store = new ReactNativeStore(this.logger)
-        } else {
-            this.store = new Store(
-                typeof window !== 'undefined' ? window.localStorage : stubbedLocalStorage, this.logger
-            )
-        }
+        this.store = new Store(
+            typeof window !== 'undefined' ? window.localStorage : stubbedLocalStorage, this.logger
```
rodeora, line 52: we shouldn't need the stubbedLocalStorage anymore

kaushalkapasi: its used for server side rendering, when there's no `window.localStorage` available to make sure the app doesn't crash

PR #343, opened 2022-12-05
Dvc 5291 read and store config cache

L-Joli:
- add cache config keys to store 
- add `disableConfigCache` and `configCacheTTL`
- set `configCacheTTL` default to 7 days
- read and store config


sdk/js/src/CacheStore.ts
```diff
@@ -1,10 +1,17 @@
 import { BucketedUserConfig, DVCLogger } from '@devcycle/types'
+import _ from 'lodash'
```
nikolasleblanc, line 2: import now from 'lodash/now'

sdk/js/src/Client.ts
```diff
@@ -54,8 +56,24 @@ export class DVCClient implements Client {
...
+        if (!this.options.configCacheTTL) {
+            this.options.configCacheTTL = 604800000
+        }
+
+        const cachedConfig = this.store.loadConfig(this.user)
+        const userId = this.store.loadConfigUserId(this.user)
+        const shouldGetConfigCache = !this.options.disableConfigCache && 
+        this.user.user_id === userId && 
+        cachedConfig &&
+        !this.isConfigCacheTTLExpired()
```
nikolasleblanc, line 69: Can this be extracted into a function with a declarative name?

sdk/js/src/Client.ts
```diff
@@ -387,6 +407,27 @@ export class DVCClient implements Client {
...
+
+    private getConfigCache(config: BucketedUserConfig) {
+        this.config = config
+        this.isConfigCached = true
+    }
+
+    private isConfigCacheTTLExpired() {
+        const cachedFetchDate = parseInt(this.store.loadConfigFetchDate(this.user) || "")
+        if (this.options.configCacheTTL) {
+            return now() - cachedFetchDate > this.options.configCacheTTL
```
nikolasleblanc, line 426:
```suggestion
            return (now() - cachedFetchDate) > this.options.configCacheTTL
```

sdk/js/src/CacheStore.ts
```diff
@@ -24,17 +31,31 @@ export abstract class CacheStore {
...
+
     remove(storeKey: string): void {
         this.store?.removeItem(storeKey)
     }
 
-    saveConfig(data: BucketedUserConfig): void {
-        this.save(StoreKey.Config, data)
+    saveConfig(data: BucketedUserConfig, user: DVCPopulatedUser): void {
+        const configKey = this.getConfigStoreKey(user)
+        this.save(configKey, data)
```
laurawarr, line 52: this method should also handle updating the fetch date and user ID

sdk/js/src/Client.ts
```diff
@@ -54,8 +56,24 @@ export class DVCClient implements Client {
 
         const storedAnonymousId = this.store.loadAnonUserId()
         this.user = new DVCPopulatedUser(user, options, undefined, storedAnonymousId ?? undefined)
-
         this.options = options
+        if (!this.options.configCacheTTL) {
+            this.options.configCacheTTL = 604800000
+        }
+
+        const cachedConfig = this.store.loadConfig(this.user)
```
laurawarr, line 64: the config shouldn't be parsed from the cache if the user ID and fetch date don't match the criteria

sdk/js/src/CacheStore.ts
```diff
@@ -24,17 +31,31 @@ export abstract class CacheStore {
...
         this.logger?.info('Successfully saved config to local storage')
     }
 
-    loadConfig(): BucketedUserConfig | null | undefined {
-        const config = this.load(StoreKey.Config)
+    loadConfig(user: DVCPopulatedUser): BucketedUserConfig | null | undefined {
+        const configKey = this.getConfigStoreKey(user)
+        const config = this.load(configKey)
```
laurawarr, line 58: ideally this method could handle validating the ID and fetch date, and only attempt to parse the config if they're valid

sdk/js/src/CacheStore.ts
```diff
@@ -60,6 +81,31 @@ export abstract class CacheStore {
...
+        this.logger?.info(`Successfully saved ${isAnonymous ? 'anonymous': 'identified'} user id to local storage`)
+    }
+
+    loadConfigUserId(user: DVCPopulatedUser): string | null | undefined {
+        const userIdKey = this.getUserIdStoreKey(user)
+        const userId = this.load(userIdKey)
+        return userId ? JSON.parse(userId) : userId
+    }
+
+    saveConfigFetchDate(user: DVCPopulatedUser): void {
```
laurawarr, line 98: the user ID and fetch date associated with the config should only be updated when the config is updated, so we shouldn't have methods for setting them individually

sdk/js/src/CacheStore.ts
```diff
@@ -1,10 +1,17 @@
 import { BucketedUserConfig, DVCLogger } from '@devcycle/types'
+import now from 'lodash/now'
```
rodeora, line 2: could use Date.now() instead of lodash

sdk/js/src/CacheStore.ts
```diff
@@ -24,17 +31,64 @@ export abstract class CacheStore {
...
+        const userIdKey = this.getUserIdStoreKey(user)
+        this.save(userIdKey, userId)
+        this.logger?.info(`Successfully saved ${isAnonymous ? 'anonymous': 'identified'} user id to local storage`)
+    }
+
+    private saveConfigFetchDate(user: DVCPopulatedUser): void {
+        const fetchDateKey = this.getConfigFetchDateKey(user)
+        this.save(fetchDateKey, now())
+        this.logger?.info('Successfully saved config fetch date to local storage')
+    }
```
rodeora, line 57: we don't need to expose these, the client should never be calling these methods directly

rodeora: yeah you can do everything in saveConfig. Also we don't have to log that we saved the config user id and fetch date, logging just the config being saved is enough

sdk/js/src/Client.ts
```diff
@@ -54,8 +55,17 @@ export class DVCClient implements Client {
...
         const storedAnonymousId = this.store.loadAnonUserId()
         this.user = new DVCPopulatedUser(user, options, undefined, storedAnonymousId ?? undefined)
-
         this.options = options
+        if (!this.options.configCacheTTL) {
+            this.options.configCacheTTL = 604800000
+        }
+        
+        if (!this.options.disableConfigCache) {
+            this.getConfigCache()
```
rodeora, line 64: Are the variable initialize callbacks going to be called if we update the config cache?

L-Joli: I'm not too sure, how can I check that?

L-Joli: added a test case to check if the `variable.value` gets the same value as the one in config cache

laurawarr: In the example app you should be able to define a client.variable.onUpdate callback outside of the client's onInitialized callback, and the onUpdate callback should be triggered with the value from the cache before the config is fetched. You may need to throttle the network connection to slow down the config fetch, or you can temporarily comment out the config fetch.

You'll likely need to look into what logic is triggered on a successful config fetch and call that from here as well.

sdk/js/src/CacheStore.ts
```diff
@@ -24,17 +30,53 @@ export abstract class CacheStore {
...
+    private getConfigStoreKey(user: DVCPopulatedUser) {
+        return user.isAnonymous ? StoreKey.AnonymousConfig : StoreKey.IdentifiedConfig
+    }
+
+    private getUserIdStoreKey(user: DVCPopulatedUser) {
+        return user.isAnonymous ? StoreKey.AnonUserId : StoreKey.IdentifiedConfigUserId
+    }
+
+    private getConfigFetchDateKey(user: DVCPopulatedUser) {
+        return user.isAnonymous ? StoreKey.AnonymousConfigDate : StoreKey.IdentifiedConfigDate
```
laurawarr, line 42: Rather than duplicating this logic, you could use the getConfigStoreKey to build the key
```suggestion
        return `${getConfigStoreKey(user)}.fetch_date`
```
This would also save us from exporting the user and date keys which shouldn't be accessible anyway

sdk/js/src/CacheStore.ts
```diff
@@ -24,17 +30,53 @@ export abstract class CacheStore {
         return this.store?.getItem(storeKey)
     }
 
+    private getConfigStoreKey(user: DVCPopulatedUser) {
+        return user.isAnonymous ? StoreKey.AnonymousConfig : StoreKey.IdentifiedConfig
+    }
+
+    private getUserIdStoreKey(user: DVCPopulatedUser) {
```
laurawarr, line 37: This key is specific to the config, so we should include that in the name. We also probably don't need to include "store" since everything in this class is scoped to the cache store
```suggestion
    private getConfigUserIdKey(user: DVCPopulatedUser) {
```

L-Joli: yup, I feel the same, not good at naming :(, thanks for the suggestion!

sdk/js/src/Client.ts
```diff
@@ -387,6 +399,23 @@ export class DVCClient implements Client {
...
         document.addEventListener?.('visibilitychange', this.pageVisibilityHandler)
     }
+
+    private cacheConfig(config: BucketedUserConfig, user: DVCPopulatedUser) {
+        this.store.saveConfig(config, user)
+        this.isConfigCached = false
+    }
+
+    private getConfigCache() {
+        if (this.options.configCacheTTL) {
```
laurawarr, line 409: `options.configCacheTTL` should be set by default, this should be skipped if `options.disableConfigCache` is set to true

sdk/js/src/Client.ts
```diff
@@ -115,6 +125,9 @@ export class DVCClient implements Client {
                 .catch((err) => onInitialized(err))
             return
         }
+        if (this.isConfigCached) {
+            this.logger.info("Initialized with a cached config")
```
laurawarr, line 129: by the time the client has initialized we would have fetched a config and set this back to false, we should log after setting the cached config on the client and triggering the onUpdate callbacks

sdk/js/src/Client.ts
L-Joli: it's skipped outside `getConfigCache()` and I've created another ticket to handle the skip config reasons
```
 if (!this.options.disableConfigCache) {
       this.getConfigCache()
} else {
       this.logger.info("Skipping config cache")
}
```

`options.configCacheTTL` is set through caz typescript does not allow ? operator and default value at the same time, not sure what's the best way to set a default value for `configCacheTTL`
```
if (!this.options.configCacheTTL) {
       this.options.configCacheTTL = 604800000
}
```

sdk/js/src/CacheStore.ts
```diff
@@ -24,17 +30,53 @@ export abstract class CacheStore {
...
+    }
+
+    private getConfigFetchDateKey(user: DVCPopulatedUser) {
+        return user.isAnonymous ? StoreKey.AnonymousConfigDate : StoreKey.IdentifiedConfigDate
+    }
+
+    private loadConfigUserId(user: DVCPopulatedUser): string | null | undefined {
+        const userIdKey = this.getUserIdStoreKey(user)
+        const userId = this.load(userIdKey)
+        return userId ? JSON.parse(userId) : userId
```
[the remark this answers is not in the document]
L-Joli: ummm...but if we do that, we'd need to assert a type in the every method we use `load` as the return type of `JSON.parse` is `any`. I'm not sure if this is a better approach

sdk/js/src/Client.ts
L-Joli: how to trigger the onUpdate callbacks?

sdk/js/src/CacheStore.ts
```diff
@@ -20,22 +22,57 @@ export abstract class CacheStore {
...
+        this.save(userIdKey, user.user_id)
         this.logger?.info('Successfully saved config to local storage')
     }
 
-    loadConfig(): BucketedUserConfig | null | undefined {
-        const config = this.load(StoreKey.Config)
-        return config ? JSON.parse(config) : config
+    loadConfig(user: DVCPopulatedUser, configCacheTTL= 604800000): BucketedUserConfig | null | undefined {
+        const userId = this.loadConfigUserId(user)
+        const cachedFetchDate = parseInt(this.loadConfigFetchDate(user) || '0')
```
laurawarr, line 68: we could move the `parseInt(... || '0')` into the loadConfigFetchDate method

laurawarr (approved):

PR #344, opened 2022-12-06
fix: only instantiate  Bucketing if it's null

nsalamad:
# Changes
* We now only instantiate `Bucketing` if it's null to allow multiple clients to exist at once


ajwootto (approved):

sdk/nodejs/src/bucketing.ts
```diff
@@ -8,8 +8,10 @@ export const importBucketingLib = async (
     { logger, options }:
     { logger?: DVCLogger, options?: DVCOptions } = {}
 ): Promise<void> => {
-    Bucketing = await instantiate()
-    startTrackingMemoryUsage(Bucketing, logger, options?.reporter)
+    if (!Bucketing) {
+        Bucketing = await instantiate()
+        startTrackingMemoryUsage(Bucketing, logger, options?.reporter)
+    }
```
jonathannorris, line 14: I'd usually return early to reduce indentation / brackets. But that's more of a style thing.

```suggestion
    if (Bucketing) return bucketing
    Bucketing = await instantiate()
    startTrackingMemoryUsage(Bucketing, logger, options?.reporter)
```

yarn.lock
```diff
@@ -8164,7 +8164,7 @@ __metadata:
 "assemblyscript-json@https://github.com/DevCycleHQ/assemblyscript-json":
   version: 1.1.0
   resolution: "assemblyscript-json@https://github.com/DevCycleHQ/assemblyscript-json.git#commit=d2ce65ec18b305e1f3db8a4fe4394b417631b024"
-  checksum: 510fc8e0828011e2257c1d4dc7fd236a8428f7f0869f59aaa0e74d8a27f9bbbc665ba74399faf466c87bcfe6fd694cd714cb130646476062bcee19c12cfc3243
+  checksum: 045c8dc54cb8ee61d022c017e483f6c8729347a2e30167a4e3227b9d4d80293b032a8cff7fcaf83040b23eacc23bf62188ebb9f2091df85ef2ee21425f8957f7
```
jonathannorris, line 8167: Why does this keep on changing?

sdk/nodejs/src/bucketing.ts
jonathannorris: Also are there any tests we should add for this?

yarn.lock
nsalamad: I don't actually know. I saw it happen in another pr and it was kept in so I figured I should keep it in. Should remove the change?

sdk/nodejs/src/bucketing.ts
nsalamad: I like that better too. I'll make the change

PR #346, opened 2022-12-07
fix: trigger test harness if nodejs files change

jsalaber:
# Summary

Fix workflow not being triggered for running test harness. Added logic to use `yarn nx print-affected` to only run test harness if `nodejs` is a project affected.

### DIFF
--- a/.github/workflows/run-test-harness.yml
+++ b/.github/workflows/run-test-harness.yml
@@ -2,8 +2,7 @@ name: Run Test Harness
 on:
   pull_request:
     branches: [main]
-    paths:
-      - 'sdk/nodejs/*'
+
 jobs:
   build:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


elliotCamblor (approved):

.github/workflows/run-test-harness.yml
```diff
@@ -3,7 +3,7 @@ on:
   pull_request:
     branches: [main]
     paths:
-      - 'sdk/nodejs/*'
+      - 'sdk/nodejs/src/*'
```
ajwootto, line 6: is this basically just only running the tests when the contents of this folder change? That's probably not a robust way to do this check because if you change any of the dependent packages of nodejs SDK (like the types package or assembly script) we'd still want to trigger the harness. You should find a way to use `nx affected` instead and run the tests any time nx detects that the nodejs sdk is "affected" by your changes.

ajwootto (approved):

PR #347, opened 2022-12-07
[feat] rewrite Cache Store to be async

kaushalkapasi:
- update DVC Client to properly use async Cache Store calls
- update tests to always `await client.onInitialized()` before interacting with the client


laurawarr (approved):
looks good to me

laurawarr:
still looks good to me, but I'll defer to @rodeora and @L-Joli since they're more familiar with the changes we've made to the js sdk recently

sdk/js/src/CacheStore.ts
```diff
@@ -17,13 +11,15 @@ export abstract class CacheStore {
         this.logger = logger
     }
 
-    private save(storeKey: string, data: unknown): void {
+    save(storeKey: string, data: unknown): void {
```
ajwootto, line 14: separate save and load into their own `Storage` class and have this class take an instance of `Storage` in its constructor

sdk/js/src/CacheStore.ts
```diff
@@ -38,46 +23,42 @@ export abstract class CacheStore {
...
 
-    private loadConfigFetchDate(user: DVCPopulatedUser): number {
+    private async loadConfigFetchDate(user: DVCPopulatedUser): Promise<number> {
         const fetchDateKey = this.getConfigFetchDateKey(user)
-        return parseInt(this.load<string>(fetchDateKey) || '0', 10)
-    }
-
-    remove(storeKey: string): void {
-        this.store?.removeItem(storeKey)
+        return parseInt(await this.store.load<string>(fetchDateKey) || '0', 10)
```
laurawarr, line 33: splitting this into two lines could make it easier to read
```suggestion
        const fetchDate = await this.store.load<string>(fetchDateKey) || '0'
        return parseInt(fetchDate, 10)
```

laurawarr (approved):

PR #350, opened 2022-12-08
add reasons for skipping config cache

L-Joli:
- add reasons for skipping config cache
1. user ID does not match
2. last fetched date is too old
3. no config found

### DIFF
--- a/sdk/js/src/CacheStore.ts
+++ b/sdk/js/src/CacheStore.ts
@@ -63,16 +63,28 @@ export abstract class CacheStore {
         this.logger?.info('Successfully saved config to local storage')
     }
 
-    loadConfig(user: DVCPopulatedUser, configCacheTTL= 604800000): BucketedUserConfig | null | undefined {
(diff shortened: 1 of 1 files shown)


sdk/js/src/CacheStore.ts
```diff
@@ -63,16 +63,28 @@ export abstract class CacheStore {
         this.logger?.info('Successfully saved config to local storage')
     }
 
-    loadConfig(user: DVCPopulatedUser, configCacheTTL= 604800000): BucketedUserConfig | null | undefined {
+    loadConfig(user: DVCPopulatedUser, configCacheTTL= 604800000): BucketedUserConfig | null {
         const userId = this.loadConfigUserId(user)
+        if (user.user_id !== userId) {
+            this.logger?.info("Skipping cached config: user ID does not match")
```
kaushalkapasi, line 69: we should set these as `debug` logs instead of `info`

kaushalkapasi (approved):

PR #352, opened 2022-12-09
update skipping config cache checking

L-Joli:
- update checking to avoid falling to unreasonable reasons
- e.g. previous logic if no config, it falls to user id not match


kaushalkapasi (approved):

sdk/js/src/Client.ts
```diff
@@ -403,8 +403,6 @@ export class DVCClient implements Client {
             this.eventEmitter.emitFeatureUpdates({}, cachedConfig.features)
             this.eventEmitter.emitVariableUpdates({}, cachedConfig.variables, this.variableDefaultMap)
             this.logger.info('Initialized with a cached config')
```
kaushalkapasi, line 405: conver to `debug` log?

PR #354, opened 2022-12-13
[feat] react native sdk

kaushalkapasi:
- RN SDK all by itself!


sdk/react-native/src/ReactNativeCacheStore.ts
```diff
@@ -0,0 +1,115 @@
...
+        const configKey = this.getConfigKey(user)
+        const fetchDateKey = this.getConfigFetchDateKey(user)
+        const userIdKey = this.getConfigUserIdKey(user)
+        this.save(configKey, data)
+        this.save(fetchDateKey, dateFetched)
+        this.save(userIdKey, user.user_id)
+        this.logger?.info('Successfully saved config to local storage')
+    }
+
+    async loadConfig(user: DVCPopulatedUser, configCacheTTL= 604800000): Promise<BucketedUserConfig | null> {
```
ajwootto, line 69: is any of this stuff platform specific or could it go in a shared base class?

sdk/react-native/src/withDVCProvider.ts
```diff
@@ -0,0 +1,17 @@
+import { withDVCProvider as reactWithDVCProvider } from '@devcycle/devcycle-react-sdk'
+import { ProviderConfig } from './types'
+import { ReactNativeStore } from './ReactNativeCacheStore'
+
+export default function withDVCProvider(config: ProviderConfig) {
```
ajwootto, line 5: you should just make this be the same type as the react provider rather than redefining ProviderConfig here

sdk/react-native/src/ReactNativeCacheStore.ts
kaushalkapasi: only the `save` / `load` / `remove` are platform specific... 🤔 the rest are the same

sdk/react-native/src/withDVCProvider.ts
```diff
@@ -0,0 +1,17 @@
...
+    const reactNativeConfig = {
+        ...config,
+        options: {
+            ...config.options,
+            reactNative: true
+        }
+    }
+    if (!config.options?.cacheStore) {
+        reactNativeConfig.options.cacheStore = new ReactNativeStore(config.options?.logger)
```
ajwootto, line 14: config.options cant be null here

sdk/react-native/src/ReactNativeCacheStore.ts
ajwootto: so can they be removed then?

sdk/react-native/src/withDVCProvider.ts
kaushalkapasi: will need to export the type from the React SDK as it currently do that

ajwootto: `typeof reactWithDVCProvider`

sdk/react-native/src/ReactNativeCacheStore.ts
kaushalkapasi: yeah, good point.. should be able to remove them :)

ajwootto: the underlying save/load methods should be in their own Storage class that is separate from all this higher-level logic that uses that Storage class..otherwise anybody providing their own implementation of storage would have to re-implement or extend all these things

sdk/react-native/src/withDVCProvider.ts
ajwootto: actually that's not true, I misread this

sdk/react-native/src/ReactNativeCacheStore.ts
kaushalkapasi: yeah, I'm updating the `DVCCacheStore` interface to just have `save`, `load` and `remove`... and then the rest is implemented by the JS SDK CacheStore... because these 3 methods will be the only ones that would be Cache Store specific implementations

ajwootto: yeah but the JS SDK `CacheStore` should basically have some pluggable `Storage` class that it calls to do `save / load`, which is the thing that this library would be replacing

sdk/react-native/src/DVCProvider.tsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react'
+import { DVCProvider as ReactDVCProvider } from '@devcycle/devcycle-react-sdk'
+import ReactNativeStore from './ReactNativeCacheStore'
+
+type configType = Parameters<typeof ReactDVCProvider>[0]
```
ajwootto, line 5: this would be more accurately named "propsType"

ajwootto: and conventionally would probably be `PropsType`

kaushalkapasi: yup! makes sense.. I was trying to make this only the `Parameters<typeof ReactDVCProvider>[0]['config']` but then didn't and forgot to rename the type back.. oops

ajwootto (approved):

PR #360, opened 2022-12-19
chore: add dvcUser class to nodejs

ajwootto:
- add separate DVCUser class that can be used to represent incoming User data
- move validation to that class
- automatically cast plain object user data to DVCUser in client methods

### DIFF
--- a/sdk/nodejs/src/client.ts
+++ b/sdk/nodejs/src/client.ts
@@ -3,8 +3,7 @@ import {
     DVCVariableValue,
     DVCVariableSet,
     DVCFeatureSet,
-    DVCEvent,
-    DVCUser
+    DVCEvent
 } from './types'
 import { EnvironmentConfigManager } from './environmentConfigManager'
 import { bucketUserForConfig } from './utils/userBucketingHelper'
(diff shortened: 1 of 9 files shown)


sdk/nodejs/src/client.ts
```diff
@@ -98,6 +105,8 @@ export class DVCClient {
     }
 
     variable<T extends DVCVariableValue>(user: DVCUser, key: string, defaultValue: T): DVCVariable<T> {
+        const incomingUser = castIncomingUser(user)
```
jsalaber, line 108: what does separating out the user into another class do for us if it does the same thing?

ajwootto: it aligns it with the other SDKs so that you can call the sdk like:

```
const user = new DVCUser({user_id: 'user'})
const variable = dvcClient.variable(user, 'key', false)
```

while maintaining backwards compatibility. Basically the other SDKs all do their user data validation inside the DVCUser class rather than within each client call, so the test harness tests can't be written the same way

jsalaber (approved):

PR #361, opened 2022-12-19
fix: nodejs should always set type field on variable instance

ajwootto:
- ensure type field is always set on DVCVariable instances in NodeJS
- field will be set to the type determined by the defaultValue passed to the variable call

### DIFF
--- a/sdk/nodejs/src/client.ts
+++ b/sdk/nodejs/src/client.ts
@@ -106,23 +106,24 @@ export class DVCClient {
 
     variable<T extends DVCVariableValue>(user: DVCUser, key: string, defaultValue: T): DVCVariable<T> {
         const incomingUser = castIncomingUser(user)
+        // this will throw if type is invalid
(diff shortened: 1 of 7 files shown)


sdk/nodejs/src/client.ts
```diff
@@ -106,23 +106,24 @@ export class DVCClient {
 
     variable<T extends DVCVariableValue>(user: DVCUser, key: string, defaultValue: T): DVCVariable<T> {
         const incomingUser = castIncomingUser(user)
+        // this will throw if type is invalid
```
jsalaber, line 109: does this mean it will throw the function as well? I thought it was supposed to log an error and just give default instead of throwing an error.

ajwootto: this is for when you've passed a default value that contains an invalid type, and yes it will throw

jsalaber (approved):

PR #362, opened 2022-12-19
fix identify user  call on JS SDK for anonymous users

kaushalkapasi:
reuses cached anonymous user id if no user_id is provided

### DIFF
--- a/sdk/js/src/Client.ts
+++ b/sdk/js/src/Client.ts
@@ -203,7 +203,8 @@ export class DVCClient implements Client {
             if (user.user_id === this.user.user_id) {
                 updatedUser = this.user.updateUser(user, this.options)
             } else {
-                updatedUser = new DVCPopulatedUser(user, this.options)
+                const storedAnonymousId = this.store.loadAnonUserId()
+                updatedUser = new DVCPopulatedUser(user, this.options, undefined, storedAnonymousId ?? undefined)
             }
 
             this.onInitialized.then(() =>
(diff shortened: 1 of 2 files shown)


sdk/js/src/Client.ts
```diff
@@ -200,10 +200,11 @@ export class DVCClient implements Client {
 
             let updatedUser: DVCPopulatedUser
 
+            const storedAnonymousId = this.store.loadAnonUserId()
```
laurawarr, line 203: we could move this to the else block, if this is only used there, that way we only read from the cache when we need to

sdk/js/src/Client.ts
```diff
@@ -200,10 +200,11 @@ export class DVCClient implements Client {
...
             if (user.user_id === this.user.user_id) {
                 updatedUser = this.user.updateUser(user, this.options)
             } else {
-                updatedUser = new DVCPopulatedUser(user, this.options)
+                updatedUser = new DVCPopulatedUser(user, this.options, undefined, storedAnonymousId ?? undefined)
```
laurawarr, line 207: is `storedAnonymousId ?? undefined` just handling the case that storedAnonymousId is null?

kaushalkapasi: yup, it is

laurawarr: no need to block this PR for this, but this seems like something we should move to `loadAnonUserId`, so that it can just return `string | undefined` rather than `string | null | undefined`

laurawarr (approved):

PR #365, opened 2022-12-21
DVC-5615 - Split the base url override into separate named overrides

JamieSinn:
Fixes the breaking baseURL override

### DIFF
--- a/sdk/nodejs/src/eventQueue.ts
+++ b/sdk/nodejs/src/eventQueue.ts
@@ -36,14 +36,14 @@ export type EventQueueOptions = {
     flushEventQueueSize?: number,
     logger: DVCLogger,
     reporter?: DVCReporter,
-    baseURLOverride?: string
+    eventsAPIBaseURLOverride?: string
 }
 
 export class EventQueue {
     private readonly logger: DVCLogger
     private readonly reporter?: DVCReporter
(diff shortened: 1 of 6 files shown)


sdk/nodejs/src/eventQueue.ts
```diff
@@ -36,14 +36,14 @@ export type EventQueueOptions = {
     flushEventQueueSize?: number,
     logger: DVCLogger,
     reporter?: DVCReporter,
-    baseURLOverride?: string
+    eventsAPIBaseURLOverride?: string
```
ajwootto, line 39: how is this making it from client initialization into here?

ajwootto: and shouldn't this just use the same name as the outer client option?

JamieSinn: Yeah - missed this in naming updates

ajwootto: it's honestly kind of weird that this class defines its own interface for the options when all of these options are supposed to come directly from the DVCOptions object..

JamieSinn: 🤷🏻

ajwootto (approved):

PR #370, opened 2023-01-04
 Always throw when the request error is a 4xx error

emmawillis:


suthar26 (approved):

sdk/nodejs/src/cloudClient.ts
```diff
@@ -61,6 +67,7 @@ export class DVCCloudClient {
                 })
             })
             .catch((err: AxiosError) => {
+                throwIfUserError(err)
```
ajwootto, line 70: do the errors being thrown here make sense to the user and indicate what they did wrong?

ajwootto, line 70: does the axios retry behaviour correctly decide to _not_ try these 4xx errors?

emmawillis: hmm not really, it just throws "AxiosError: Request failed with status code 401"
I thought that was all we got back from the server but the response.message actually has useful stuff like "Error: Invalid SDK token provided" and "Error: Invalid User Data to process request" so i updated it to throw that

emmawillis: our retry condition is 
```
retryCondition: (error) => {
        return !error.response || (error.response.status || 0) >= 500
    }
```
so it shouldn't retry 400 errors

sdk/nodejs/src/cloudClient.ts
```diff
@@ -22,6 +22,14 @@ const castIncomingUser = (user: DVCUser) => {
     return user
 }
 
+const throwIfUserError = (err: AxiosError) => {
+    const data = err.response?.data as { message: string, statusCode: number }
+
+    if (400 <= data.statusCode && data.statusCode < 500) {
```
ajwootto, line 28: you should read the status code of the response object here, not the data. Data is our response body from the server, which for some reason contains a status code that is the same as the one on the response itself. Just use `response.status` or whatever it is

sdk/nodejs/__tests__/cloudClient.spec.ts
```diff
@@ -251,9 +254,10 @@ describe('DVCCloudClient with EdgeDB Enabled', () => {
             expect(res).toEqual({})
         })
 
-        it('to return no features when allFeatures call fails', async () => {
```
ajwootto, line 254: seems like this test should still exist? its testing a different case than the invalid user one

sdk/nodejs/__tests__/cloudClient.spec.ts
```diff
@@ -251,9 +254,10 @@ describe('DVCCloudClient with EdgeDB Enabled', () => {
...
-            const res = await client.allFeatures(badUser)
-            expect(res).toEqual({})
+        it('throws exception when allFeatures called with invalid user', async () => {
+            await expect(async () => {
+                await client.allFeatures(badUser)
+            }).rejects.toThrow('Bucketing request failed with status 401.')
```
ajwootto, line 260: this shouldnt be a 401 if the only problem is an invalid user

sdk/nodejs/src/__mocks__/handlers.ts
```diff
@@ -4,7 +4,11 @@ import { rest } from 'msw'
 export const handlers = [
     rest.post('https://bucketing-api.devcycle.com/v1/variables/test-key-not-in-config', (req, res, ctx) => {
         return res(
-            ctx.status(400)
+            ctx.status(404),
+            ctx.json({
+                statusCode: 404,
+                message: 'Feature not found for key: test-key-not-in-config'
```
ajwootto, line 10: variable?

sdk/nodejs/src/cloudClient.ts
```diff
@@ -22,6 +22,15 @@ const castIncomingUser = (user: DVCUser) => {
     return user
 }
 
+const throwIfUserError = (err: AxiosError, isVariableRequst?: boolean) => {
+    const data = err.response?.data as { message: string, statusCode: number }
```
ajwootto, line 26: no need to define the `statusCode` field here since you aren't using it 

I'd also make the message field optional since you dont know for sure that it's actually there.

sdk/nodejs/src/__mocks__/handlers.ts
emmawillis: oh yeah i fixed this in the actual api, it doesn't get used anymore here so i'll just take it out the the mock

sdk/nodejs/src/cloudClient.ts
```diff
@@ -22,6 +22,15 @@ const castIncomingUser = (user: DVCUser) => {
     return user
 }
 
+const throwIfUserError = (err: AxiosError, isVariableRequst?: boolean) => {
```
ajwootto, line 25: is the `isVariableRequest` parameter being used?

yarn.lock
```diff
@@ -8202,7 +8202,7 @@ __metadata:
 "assemblyscript-json@https://github.com/DevCycleHQ/assemblyscript-json":
   version: 1.1.0
   resolution: "assemblyscript-json@https://github.com/DevCycleHQ/assemblyscript-json.git#commit=d2ce65ec18b305e1f3db8a4fe4394b417631b024"
-  checksum: 045c8dc54cb8ee61d022c017e483f6c8729347a2e30167a4e3227b9d4d80293b032a8cff7fcaf83040b23eacc23bf62188ebb9f2091df85ef2ee21425f8957f7
+  checksum: 7485348b948af8a6c8dde91714d09dd1d966597a670fb3d4d484135a58151f17b71caaf423c0d2b46869cc9d06d0febb10e69064c849f501d1396f071369a82b
```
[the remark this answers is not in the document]
ajwootto: should be fine, we gotta figure this thing out though with that guild ticket

sdk/nodejs/src/cloudClient.ts
```diff
@@ -22,6 +22,15 @@ const castIncomingUser = (user: DVCUser) => {
...
+    const data = err.response?.data as { message: string, statusCode: number }
+    const code = err.response?.status || 0
+
+    if (code !== 404 && code < 500) {
+        throw new Error(`Bucketing request failed with status ${code}. ${data.message || ''}`)
```
ajwootto, line 30: this error message isn't accurate in the case where you called Track or something, which isn't really a "bucketing" request. Maybe this could just say `DevCycle request failed` or something, unless you want to include the name of the actual method that was being called, like `allVariables Request failed`

sdk/nodejs/__tests__/cloudClient.spec.ts
emmawillis: this is just because the mock server returns 401 when the user id is 'bad user', but the actual api returns 400 so i guess it makes sense to match

sdk/nodejs/src/cloudClient.ts
emmawillis: oh true, fixed!

sdk/nodejs/src/cloudClient.ts
emmawillis: good point, i'll use `DevCycle request failed` since the method should be in the stack trace

sdk/nodejs/__tests__/cloudClient.spec.ts
emmawillis: the tests time out for 500 level errors because it retries and we don't have an option to turn off retries from the client level, so now since the 400s throw we don't have a good way to test this (other than having the test timeout set to 10+ seconds)

sdk/nodejs/src/cloudClient.ts
emmawillis: nope, my bad - fixed!

ajwootto (approved):

PR #385, opened 2023-01-30
feat: added class for user error to prevent swallowing of errors to throw

jsalaber:
# Summary

Added a `UserError` class to distinguish between errors to be swallowed and errors to be thrown back to the user

### DIFF
--- a/sdk/nodejs/src/client.ts
+++ b/sdk/nodejs/src/client.ts
@@ -17,6 +17,7 @@ import { importBucketingLib, getBucketingLib } from './bucketing'
 import { DVCLogger, getVariableTypeFromValue, VariableTypeAlias } from '@devcycle/types'
 import os from 'os'
 import { DVCUser } from './models/user'
+import { UserError } from './utils/userError'
 
 interface IPlatformData {
     platform: string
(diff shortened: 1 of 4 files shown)


sdk/nodejs/src/client.ts
```diff
@@ -50,6 +51,9 @@ export class DVCClient {
         }
 
         const initializePromise = importBucketingLib({ options, logger: this.logger })
+            .catch((bucketingErr) => {
+                throw new UserError(`Error importing bucketing library: ${bucketingErr}`)
```
ajwootto, line 55: does this obscure the stack trace of the original error?

jsalaber: added support for creating it from `Error` objects

ajwootto: you'd have to pass the error in here then

sdk/nodejs/src/utils/userError.ts
```diff
@@ -0,0 +1,7 @@
+export class UserError extends Error {
+    constructor(error: Error | string) {
+        super(error instanceof Error ? error.message : error)
+        this.name = 'UserError'
+        this.stack = error instanceof Error ? error.stack : undefined
```
ajwootto, line 5: does this make it so the stack isnt set properly when it's thrown with a string message?

jsalaber: this is for the cases where we throw an error when we know something is wrong and we throw with a string message

ajwootto (approved):

PR #386, opened 2023-01-31
fix: calling flushEvents multiple times should result in one flush

nsalamad:
# Changes
* Calling `flushEvents` multiple times while a flush is in progress will now only result in a single flush
* `checkEventQueueSize` no longer needs to check if a flush is in progress and can just call `flushEvents`

**Note**: put on hold for now

### DIFF
--- a/sdk/nodejs/src/eventQueue.ts
+++ b/sdk/nodejs/src/eventQueue.ts
@@ -49,6 +49,7 @@ export class EventQueue {
     maxEventQueueSize: number
     private flushInterval: NodeJS.Timer
     private flushInProgress = false
+    private flushCallbacks: Array<(arg: unknown) => void> = []
 
     constructor(sdkKey: string, options: EventQueueOptions) {
         this.logger = options.logger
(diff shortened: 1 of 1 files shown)


ajwootto (approved):

elliotCamblor (approved):

sdk/nodejs/src/eventQueue.ts
```diff
@@ -117,7 +124,12 @@ export class EventQueue {
         const flushPayloads = JSON.parse(flushPayloadsStr) as FlushPayload[]
         const endTimeJson = Date.now()
         this.reporter?.reportMetric('jsonParseDuration', endTimeJson - startTimeJson, metricTags)
-        if (flushPayloads.length === 0) return
+        if (flushPayloads.length === 0) {
+            if (currentFlushCallbacks.length > 0) {
+                currentFlushCallbacks.forEach((cb) => cb(results))
```
ajwootto, line 129: this should probably be called with nothing like the one below

ajwootto (approved):

PR #394, opened 2023-02-03
update bucketing lib for reusable audiences

elliotCamblor:


lib/shared/bucketing-assembly-script/__tests__/bucketing/segmentation.test.ts
```diff
@@ -394,12 +394,52 @@ describe('SegmentationManager Unit Test', () => {
...
+                }
+                assert.strictEqual(true, evaluateOperator({ data, operator: audienceMatchOperator }))
+            })
+        })
+
         describe('evaluateOperator should handle a new user sub-filter (myNewFilter) type', () => {
             const filters = [{
                 type: 'user',
-                subType: 'myNewFilter',
+                subType: 'email',
```
ajwootto, line 440: whats this change for?

lib/shared/bucketing-assembly-script/__tests__/bucketing/segmentation.test.ts
```diff
@@ -394,12 +394,52 @@ describe('SegmentationManager Unit Test', () => {
...
+                    platform: 'iOS'
+                }
+
+                const audienceMatchOperator = {
+                    filters: [
+                        {
+                          type: 'audienceMatch',
+                            comparator: '=',
+                          _audiences: [
+                              {
```
ajwootto, line 424: these arent going to be populated in the config so you should test it when its just a string id

ajwootto: oh unless youre populating it before calling evaluateOperator

lib/shared/bucketing-assembly-script/assembly/bucketing/segmentation.ts
```diff
@@ -74,6 +79,24 @@ function doesUserPassFilter(
...
+    // Recursively evaluate every audience in the _audiences array
+    for (let i = 0; i < audiences.length; i++) {
+        const audience = configAudiences.get(audiences[i])
+        if (_evaluateOperator(audience.filters, user, configAudiences)) {
+            // If the user is in any of the audiences return early.
+            return comparator === '='
+        }
+    }
+    // The user is not in any of the audiences.
+    return comparator !== '='
```
ajwootto, line 98: you should probably be explicitly checking for `!=` as the comparator instead of doing "anything except '='"

lib/shared/bucketing-assembly-script/__tests__/bucketing/segmentation.test.ts
elliotCamblor: I was originally which is why it was this way in the tests. Its changed now though

lib/shared/bucketing-assembly-script/assembly/bucketing/segmentation.ts
```diff
@@ -314,6 +337,29 @@ export function _checkCustomData(data: JSON.Obj | null, filter: CustomDataFilter
     }
 }
 
+export function getFilterAudiences(filter: AudienceMatchFilter): JSON.Value[] {
+    const _audiences = filter._audiences
+
+    return _audiences.valueOf().reduce((accumulator, audience) => {
```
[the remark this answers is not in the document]
elliotCamblor: This is how its done for the values field, I just borrowed the same approach

lib/shared/bucketing-assembly-script/__tests__/bucketing/segmentation.test.ts
```diff
@@ -394,6 +395,184 @@ describe('SegmentationManager Unit Test', () => {
...
+                    },
+                    'nested': {
+                        _id: 'nested',
+                        filters: audienceMatchOperator
+                    }
+                }
+                assert.strictEqual(false, evaluateOperator(
+                    { data, operator: nestedAudienceMatchOperator, audiences }))
+            })
+            it('should pass seg for an AND operator with multiple values', () => {
```
ajwootto, line 509: i dont think the and operator is really being tested in this test, since it applies to the "filters" array and all your filters arrays only have a length of 1

lib/shared/bucketing-assembly-script/__tests__/bucketing/segmentation.test.ts
```diff
@@ -394,6 +395,184 @@ describe('SegmentationManager Unit Test', () => {
             })
         })
 
+        describe('evaluateOperator should handle audienceMatch filter', () => {
```
ajwootto, line 398: it would be good to also have at least one test that tests this filter from the entry point of segmentation where the config data is passed. Currently we aren't covering the earlier code paths that pass the audience through to `evaluateOperator`

lib/shared/bucketing-assembly-script/__tests__/bucketing/segmentation.test.ts
elliotCamblor: yeah I will rename this test. Its just titled like this because AND operator is the default. I dont think we need to re-write tests for the foundational segmentation code for every new filter type. The code that handles the AND level filtering hasnt changed at all, so as long as the code that actually tests the `audieceMatch` filter type works then we know that AND filtering on it will also work.

ajwootto: well don't we need to test that audiences are correctly passed down?

lib/shared/bucketing-assembly-script/__tests__/bucketing/segmentation.test.ts
elliotCamblor: yeah good callout, ill add a test in the bucketing.test file that tests the output of a bucketed config

lib/shared/bucketing-assembly-script/__tests__/bucketing/segmentation.test.ts
elliotCamblor: The AND operator functionality is contained inside the function that all these tests are calling, so that's already tested. Plus it will be double covered once I add a test to the bucketing file that calls the parent function of that

ajwootto: oh sorry i thought this was in reference to the other comment about testing at a higher level

ajwootto:
* main
  * **PR #394** <a href="https://app.graphite.dev/github/pr/DevCycleHQ/js-sdks/394" target="_blank">Graphite</a>  👈
    * **PR #399** <a href="https://app.graphite.dev/github/pr/DevCycleHQ/js-sdks/399" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/DevCycleHQ/js-sdks/394?utm_source=stack-comment).

lib/shared/bucketing-test-data/src/data/testData.ts
```diff
@@ -248,7 +275,14 @@ export const variableHashes: ConfigBody['variableHashes'] = {
 export const config: ConfigBody = {
     project,
     environment,
-    audiences: {},
+    audiences: {
+        [audiences[0]._id]: audiences[0],
```
ajwootto, line 279: you should omit the _ids from these to match what Nika is doing with the config

lib/shared/bucketing-assembly-script/assembly/bucketing/segmentation.ts
```diff
@@ -18,18 +19,20 @@ import { getF64FromJSONValue } from '../helpers/jsonHelpers'
  * Returns true if the user's data allows them through the segmentation
  * @param operator - The set of filters to evaluate, and the boolean operator to follow (AND, OR, XOR)
  * @param user - The incoming user, device, and user agent data
+ * @param audiences - a map of audience_id to audience, used for audienceMatch filters
  */
 export function _evaluateOperator(
     operator: TopLevelOperator,
-    user: DVCPopulatedUser
+    user: DVCPopulatedUser,
+    audiences: Map<string, Audience>
```
ajwootto, line 27: if we follow Nika's config stuff, this would technically not have an _id field in the Audience

lib/shared/bucketing-test-data/src/data/testData.ts
```diff
@@ -245,10 +272,19 @@ export const variableHashes: ConfigBody['variableHashes'] = {
     'feature2.hello': 4138596111
 }
 
+function configBodyAudiences (audiences: PublicAudience[]): { [id: string]: Omit<PublicAudience<string>, '_id'> }{
+    const auds: {[id: string]: Omit<PublicAudience<string>, "_id">} = {}
+    audiences.forEach((aud: PublicAudience) => {
+        auds[aud._id] = {
+            ...aud
```
ajwootto, line 279: dont you have to actually remove the id here?

ajwootto (approved):

PR #401, opened 2023-02-07
feat: add client custom data to bucketing

ajwootto:


ajwootto:
* main
  * **PR #401** <a href="https://app.graphite.dev/github/pr/DevCycleHQ/js-sdks/401" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/DevCycleHQ/js-sdks/401?utm_source=stack-comment).

lib/shared/bucketing-assembly-script/__tests__/bucketing/bucketing.test.ts
```diff
@@ -644,3 +649,57 @@ describe('Rollout Logic', () => {
...
+            platformVersion: '1.1.2'
+        }
+
+        const c3 = generateBucketedConfig({ config, user: user2 })
+        expect(c3).toEqual(expect.objectContaining({
+            'featureVariationMap': {}
+        }))
+
+        // cleanup
+        setClientCustomDataJSON({})
```
elliotCamblor, line 703: I would be in favour of having an actual cleanup method that we could use in the `close` method

lib/shared/bucketing-assembly-script/assembly/index.ts
```diff
@@ -6,6 +6,7 @@ import { JSON } from 'assemblyscript-json/assembly'
 import { ConfigBody, DVCPopulatedUser, PlatformData } from './types'
 import { _clearPlatformData, _setPlatformData } from './managers/platformDataManager'
 import { _getConfigData, _setConfigData } from './managers/configDataManager'
+import { _getClientCustomData, _setClientCustomData } from './managers/globalDataManager'
```
elliotCamblor, line 9: why are you calling it `clientCustomData` in some places but calling it `globalData` in others? Do you forsee the concept of shared data going beyond `client` scoped in the context of wasm?

lib/shared/bucketing-assembly-script/__tests__/bucketing/bucketing.test.ts
ajwootto: Well this was just meant to clean up the test's side effects so it wouldn't interfere with other tests in this file.

I can see a need to clean this up when a client is closed, but it would be similar to cleaning up the config data (which we also don't do right now).
Maybe it's better to tackle all of that at once, in a single "clean up this client from the WASM" type operation like you were working on before?

elliotCamblor: yeah I think that it would be fine to tackle this separately as part of a larger cleanup task

elliotCamblor (approved):

PR #403, opened 2023-02-10
Add nested filter support in bucketing wasm

elliotCamblor:


lib/shared/bucketing-assembly-script/assembly/bucketing/segmentation.ts
```diff
@@ -34,7 +34,10 @@ export function _evaluateOperator(
         // Replace Array.some() logic
         for (let i = 0; i < operator.filters.length; i++) {
             const filter = operator.filters[i]
-            if (doesUserPassFilter(filter, audiences, user, clientCustomData)) {
+            if (filter.operatorClass !== null) {
+                return _evaluateOperator(filter.operatorClass as AudienceOperator, audiences, user, clientCustomData)
```
ajwootto, line 38: is the type assertion here just to remove the "null"?

lib/shared/bucketing-assembly-script/assembly/types/target.ts
```diff
@@ -43,65 +43,112 @@ export class Target extends JSON.Value {
...
+// The config is returned to its original form when stringified.
+export class AudienceFilterOrOperator extends JSON.Value {
+    operatorClass: AudienceOperator | null
+    filterClass: AudienceFilter | null
+
+    constructor(filter: JSON.Obj) {
+        super()
+        const operator = isValidStringOptional(filter, 'operator', validAudienceOperators)
+        this.operatorClass = operator ? new AudienceOperator(filter) : null
+        this.filterClass = initializeFilterClass(filter)
```
ajwootto, line 74: this should throw if neither of these get set

ajwootto: I guess it always tries to turn it into a filter class, and that will fail

lib/shared/bucketing-test-data/src/data/testData.ts
```diff
@@ -49,6 +49,20 @@ export const audiences: PublicAudience[] = [
         _id: '6153557f1ed7bac7268ea0d9',
         filters: {
             filters: [{
+                filters: [{
```
ajwootto, line 52: any need for a more targeted segmentation test that just unit tests the nested logic?

lib/shared/bucketing-assembly-script/assembly/bucketing/segmentation.ts
elliotCamblor: yep

lib/shared/bucketing-test-data/src/data/testData.ts
elliotCamblor: I think it couldnt hurt to add one, so I will

ajwootto: it's a lot easier to represent all the different cases the code is supposed to handle as targeted unit tests, rather than burying it implicitly in the test data. Also makes it clearer to see what went wrong when something fails

lib/shared/bucketing-assembly-script/assembly/types/target.ts
elliotCamblor: it actually wasnt failing in the `initializeFilterClass` function, so I changed that

lib/shared/bucketing-test-data/src/data/testData.ts
elliotCamblor: yeah 100p

elliotCamblor: Added a bunch of segmentation level tests

lib/shared/bucketing-assembly-script/assembly/types/target.ts
```diff
@@ -231,18 +234,46 @@ export class CustomDataFilter extends UserFilter {
...
             return new CustomDataFilter(filter)
         }
         return new UserFilter(filter)
     } else if (getStringFromJSONOptional(filter, 'type') === 'audienceMatch'){
         return new AudienceMatchFilter(filter)
+    } else if (getStringFromJSONOptional(filter, 'type') === 'all'){
+        return new AudienceFilter(filter)
     } else {
-        return new AudienceFilterOrOperator(filter)
+        throw new Error('Invalid target filter structure. Empty filters detected.')
```
ajwootto, line 276: wont this now throw whenever a new type is added that isnt covered by the if statement above?

elliotCamblor: doesnt that contradict your earlier comment about how the AudienceOperatorOrFilter class should throw if filterClass and operatorClass are both null? If I dont do it this way then it will always produce a filter and never throw

ajwootto: well what it used to do is generate a valid "AudienceFilter" class as long as there was a type of any kind set. I was more thinking that it should basically be impossible for both those fields to be null, either you'd have an operator or an audience filter set. But the audience filter might be "invalid".

We have downstream code which handles for audience filters that have an unknown type, which is why it needs to be initialized that way

elliotCamblor: yeah, so basically its impossible to throw in this class if theyre both null, because if the filter is ever null that means we arent forwards compatible in adding new types

ajwootto: yeah but the filter shouldnt be null if the type is unknown, it should be `AudienceFilter`

lib/shared/bucketing-assembly-script/assembly/types/target.ts
```diff
@@ -43,65 +43,112 @@ export class Target extends JSON.Value {
...
+// This is a virtual container class that's used by the code to determine whether a filter is a nested operator
+// or a base level filter. This is used to support the recursive nature of nested filters.
+// The config is returned to its original form when stringified.
+export class AudienceFilterOrOperator extends JSON.Value {
+    operatorClass: AudienceOperator | null
+    filterClass: AudienceFilter | null
+
+    constructor(filter: JSON.Obj) {
+        super()
+        const operator = isValidStringOptional(filter, 'operator', validAudienceOperators)
```
ajwootto, line 72: this should be passing false to "shouldThrow".

might need a test to cover this

elliotCamblor: it actually shouldnt because its using `isValidStringOptional`, so it wont throw

ajwootto: afaik the Optional version only doesnt throw when the field is actually null, if it's set to a value that isnt in the set of valid types it throws

lib/shared/bucketing-assembly-script/assembly/types/target.ts
```diff
@@ -231,9 +234,35 @@ export class CustomDataFilter extends UserFilter {
...
+        if (this.values) {
+            json.set('values', this.values)
+        }
+        return json.stringify()
+    }
 }
 
-function initializeFilterClass(filter: JSON.Obj): AudienceFilterOrOperator {
+function initializeFilterClass(filter: JSON.Obj): AudienceFilter | null {
+    if (isValidStringOptional(filter, 'operator', validAudienceOperators, false)){
```
ajwootto, line 262: dont think we want this here anymore?

elliotCamblor: uhhh yeah, I can change so that this method doesn't get called if it's an operator instead. It makes more sense there anyway

jsalaber (approved):

lib/shared/bucketing-assembly-script/assembly/types/target.ts
```diff
@@ -231,9 +234,31 @@ export class CustomDataFilter extends UserFilter {
...
+        }
+        if (this.values) {
+            json.set('values', this.values)
+        }
+        return json.stringify()
+    }
 }
 
-function initializeFilterClass(filter: JSON.Obj): AudienceFilterOrOperator {
+function initializeFilterClass(filter: JSON.Obj): AudienceFilter | null {
```
ajwootto, line 261: this cant return null

elliotCamblor: fffffffffixed I mean

ajwootto (approved):

PR #409, opened 2023-02-21
DVC-6240 prefix env keys with NX

nikolasleblanc:


jonathannorris (approved):

examples/js/web-elements-app/src/app/app.element.ts
```diff
@@ -3,7 +3,7 @@
 import './app.element.css'
 import { initialize } from '@devcycle/devcycle-js-sdk'
 
-const SDK_KEY = process.env['DVC_CLIENT_SDK_KEY'] || '<YOUR_DVC_CLIENT_SDK_KEY>'
+const SDK_KEY = process.env.NX_DVC_CLIENT_SDK_KEY || '<YOUR_DVC_CLIENT_SDK_KEY>'
```
jonathannorris, line 6: do we need to do this for the non NX example apps?

PR #413, opened 2023-03-03
DVC-6473 add variableForUser() method to WASM code

jonathannorris:
Add `variableForUser()` method in the WASM code that will return a specific variable result for a user, rather than fetching the full `generateBucketedConfigForUser()` every time. Also generate `aggVariableEvaluated` and `aggVariableDefaulted` from within the WASM code.

PR #414, opened 2023-03-04
DVC-6489 Efficient Variable Evaluation

ajwootto:
- updating WASM code to efficiently evaluate variables

### DIFF
--- a/lib/shared/bucketing-assembly-script/assembly/types/configBody.ts
+++ b/lib/shared/bucketing-assembly-script/assembly/types/configBody.ts
@@ -86,6 +86,11 @@ export class ConfigBody {
     variables: Variable[]
     variableHashes: Map<string, i64>
 
+    private _variableKeyMap: Map<string, Variable>
+
+    private _varIdToFeatureMap: Map<string, Feature>
+
+
     constructor(configStr: string) {
         const configJSON = JSON.parse(configStr)
         if (!configJSON.isObj) throw new Error('generateBucketedConfig config param not a JSON Object')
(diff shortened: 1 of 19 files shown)


ajwootto:
* main
  * **PR #413** <a href="https://app.graphite.dev/github/pr/DevCycleHQ/js-sdks/413" target="_blank">Graphite</a> 
    * **PR #414** <a href="https://app.graphite.dev/github/pr/DevCycleHQ/js-sdks/414" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/DevCycleHQ/js-sdks/414?utm_source=stack-comment).

lib/shared/bucketing-assembly-script/assembly/types/configBody.ts
```diff
@@ -111,10 +129,19 @@ export class ConfigBody {
...
-        })
+        const variablesJSON = getJSONArrayFromJSON(configJSONObj, 'variables').valueOf()
+        const variables = new Array<Variable>()
+        const _variableKeyMap = new Map<string, Variable>()
+        const _variableIdMap = new Map<string, Variable>()
+        for (let i=0; i < variablesJSON.length; i++) {
+            const variable = new Variable(variablesJSON[i] as JSON.Obj)
+            variables.push(variable)
+            _variableKeyMap.set(variable.key, variable)
+            _variableIdMap.set(variable._id, variable)
```
elliotCamblor, line 140: do we actually need the `_variableIdMap` at the config level? I only see the `_variableKeyMap` being used.

If we do need it, I think it would be slightly more space efficient (and practically the same time efficiency) to have a `_variableIdKey` map that holds the _id to key mapping. That way we dont need to store every variable twice at the config level.

ajwootto: in theory the two should be equivalent because its not storing the variable in each one, its just storing a reference to it

jonathannorris: yea technically we need both, `_variableIdMap` is used here: https://github.com/DevCycleHQ/js-sdks/pull/414/files#diff-e3b505010bb193bc1aedac8338f00363231ff4b24b565eb0989f678b2e772962R279

and the `_variableKeyMap` is used here: https://github.com/DevCycleHQ/js-sdks/pull/414/files#diff-e3b505010bb193bc1aedac8338f00363231ff4b24b565eb0989f678b2e772962R325

jonathannorris (approved):

PR #416, opened 2023-03-06
chore: include hostname in dvcpopulateduser

jsalaber:
# Summary

- Include `hostname` in `DVCPopulatedUser`


elliotCamblor (approved):

lib/shared/bucketing-assembly-script/assembly/types/dvcUser.ts
```diff
@@ -97,6 +97,7 @@ export class DVCPopulatedUser extends JSON.Value implements DVCUserInterface {
     platformVersion: string
     sdkType: string
     sdkVersion: string
+    hostname: string | null
```
L-Joli, line 100: when'd it be null?

jsalaber: in the case where `os.hostname` returns a null, since we can't guarantee it will give a value as opposed to a hard coded value

PR #418, opened 2023-03-06
DVC-6504 update variableType param in variableForUser() to be an enum

jonathannorris:
- update variableType param in `variableForUser()` to be an enum


ajwootto (approved):

ajwootto (approved):

nikolasleblanc (approved):

jonathannorris:
@nikolasleblanc yea basically allocating strings is slow, so if we can combine multiple allocations into one or use ints/bools then we get a performance win.

PR #419, opened 2023-03-07
DVC-6484 add hasConfigData() WASM method

jonathannorris:
- `setConfigDataWithEtag()` and `hasConfigDataForEtag()` methods will be used to check if we already have a config set for a specific etag value. This will mainly be used in our CF workers to optimize how often we set the config.

### DIFF
--- a/lib/shared/bucketing-assembly-script/__tests__/types/configBody.test.ts
+++ b/lib/shared/bucketing-assembly-script/__tests__/types/configBody.test.ts
@@ -1,5 +1,9 @@
 import testData from '../../../bucketing-test-data/json-data/testData.json'
-import { testConfigBodyClass } from '../bucketingImportHelper'
+import {
+    testConfigBodyClass,
+    setConfigData,
+    hasConfigDataForEtag
+} from '../bucketingImportHelper'
 import cloneDeep from 'lodash/cloneDeep'
 
 describe('Config Body', () => {
@@ -114,5 +118,16 @@ describe('Config Body', () => {
         for (let i = 0; i < 1000; i++) {
             testConfigBodyClass(JSON.stringify(testData.config))
         }
+    })
+})
+
+describe('hasConfigDataForEtag', () => {
+    it('should return true if config has data', () => {
+        const sdkKey = 'sdkKey'
+        const etag = 'etag'
+        setConfigData(sdkKey, JSON.stringify(testData.config), etag)
+        expect(hasConfigDataForEtag(sdkKey, etag)).toBe(true)
+        expect(hasConfigDataForEtag(sdkKey, 'not-etag')).toBe(false)
+        expect(hasConfigDataForEtag('not-sdk-key', etag)).toBe(false)
     })
 })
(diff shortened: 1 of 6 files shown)


sdk/nodejs/src/environmentConfigManager.ts
```diff
@@ -103,9 +103,10 @@ export class EnvironmentConfigManager {
             return
         } else if (res?.status === 200 && projectConfig) {
             try {
+                const etag = res?.headers.get('etag') || ''
                 getBucketingLib().setConfigData(this.sdkKey, projectConfig)
```
ajwootto, line 107: were you intending to pass it in here?

jonathannorris: I was during testing, but need to wait to release the WASM code first b4 I can update the NodeJS SDK as its WASM version is pinned.

lib/shared/bucketing-assembly-script/__tests__/types/configBody.test.ts
```diff
@@ -1,5 +1,10 @@
 import testData from '../../../bucketing-test-data/json-data/testData.json'
-import { testConfigBodyClass } from '../bucketingImportHelper'
+import {
+    testConfigBodyClass,
+    setConfigData,
```
nikolasleblanc, line 4: Should this be removed then?

lib/shared/bucketing-assembly-script/assembly/testHelpers.ts
```diff
@@ -97,8 +97,8 @@ export function doesUserPassRolloutFromJSON(rolloutStr: string | null, boundedHa
     return _doesUserPassRollout(rollout, boundedHash)
 }
 
-export function testConfigBodyClass(configStr: string): string {
-    const config = new ConfigBody(configStr)
+export function testConfigBodyClass(configStr: string, etag: string | null = null): string {
+    const config = new ConfigBody(configStr, etag)
```
nikolasleblanc, line 101: Could this be...

```suggestion
export function testConfigBodyClass(configStr: string, etag?: string): string {
    const config = new ConfigBody(configStr, etag || null)
```

jonathannorris: No AS doesn't allow direct optionals like that, you need to use `| null` instead of `?`, and I prefer putting the default values in the param definition.

lib/shared/bucketing-assembly-script/assembly/managers/configDataManager.ts
```diff
@@ -13,3 +13,7 @@ export function _getConfigData(sdkKey: string): ConfigBody {
         return _configData.get(sdkKey)
     }
 }
+
+export function _hasConfigData(sdkKey: string): bool {
+    return _configData.has(sdkKey)
+}
```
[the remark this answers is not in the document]
jonathannorris: I think because `_configData` is private we can't do that.

lib/shared/bucketing-assembly-script/__tests__/types/configBody.test.ts
jonathannorris: no that's still used by all the existing tests. Just added a new test for the new methods.

ajwootto (approved):

PR #427, opened 2023-03-08
DVC-6519 protobuf interface for variableForUser

jonathannorris:
- adding `variableForUser_PB()` method that supports accepting params as a Protobuf, and returning a `SDKVariable` as a protobuf.


lib/shared/bucketing-assembly-script/assembly/types/protobuf/as-generated/DVCUser_PB.ts
```diff
@@ -0,0 +1,156 @@
+// Code generated by protoc-gen-as. DO NOT EDIT.
```
jonathannorris, line 1: Do we want to include these generated files? or should we gitignore them?

jonathannorris: thoughts @ajwootto ?

chris-hoefgen (approved):

lib/shared/bucketing-assembly-script/assembly/types/dvcUser.ts
```diff
@@ -33,6 +65,44 @@ export class DVCUser extends JSON.Obj implements DVCUserInterface {
...
+            (nullableAppBuild && !nullableAppBuild.isNull) ? nullableAppBuild.value : NaN,
+            (nullableAppVersion && !nullableAppVersion.isNull) ? nullableAppVersion.value : null,
+            (nullableDeviceModel && !nullableDeviceModel.isNull) ? nullableDeviceModel.value : null,
+            getJSONObjFromPBCustomData(nullableCustomData),
+            getJSONObjFromPBCustomData(nullablePrivateCustomData),
+        )
+    }
+
+    toProtoBuf(): Uint8Array {
+        return encodeDVCUser_PB(new DVCUser_PB(
```
ajwootto, line 94: can we not directly use this DVCUser_PB object rather than translating back and forth to our existing dvcUser class? or is that too hard to integrate with the existing code?

I have a feeling we'd gain even more by not having to copy all this memory every time

lib/shared/bucketing-assembly-script/assembly/types/dvcUser.ts
```diff
@@ -17,6 +26,29 @@ interface DVCUserInterface {
...
+        const key = keys[i]
+        const value: CustomDataValue = nullableCustomData.value.get(key)
+        if (value && value.type === VariableType_PB.Boolean) {
+            customDataObj.set(key, value.boolValue)
+        } else if (value && value.type === VariableType_PB.Number) {
+            customDataObj.set(key, value.doubleValue)
+        } else if (value && value.type === VariableType_PB.String) {
+            customDataObj.set(key, value.stringValue)
+        } else {
+            throw new Error('DVCUser customData can\'t contain nested objects or arrays')
```
ajwootto, line 45: what about null values?

JamieSinn: Nulls should be allowed - see current client-sdk not accepting nulls.

lib/shared/bucketing-assembly-script/assembly/types/protobuf/as-generated/DVCUser_PB.ts
ajwootto: not fussed either way, probably makes sense to include them since otherwise you'd get type errors when cloning the repo until you build it

lib/shared/bucketing-assembly-script/assembly/types/dvcUser.ts
chris-hoefgen: Potentially yes, but I think it would require some extensive changes since we had to do a lot of munging to fit these objects into how protobuf can represent them

lib/shared/bucketing-assembly-script/assembly/types/dvcUser.ts
ajwootto: correct. we want to allow explicitly setting a key's value to null

lib/shared/bucketing-assembly-script/assembly/types/dvcUser.ts
jonathannorris: yea protobuf can't really do nulls easily, so we need these translation layers. We can look into optimizations later, but I don't think they will help much.

lib/shared/bucketing-assembly-script/assembly/types/dvcUser.ts
jonathannorris: that might be a little tricky to add, will take a look

lib/shared/bucketing-assembly-script/assembly/types/dvcUser.ts
jonathannorris: also this method is only used for testing.

lib/shared/bucketing-assembly-script/assembly/types/dvcUser.ts
jonathannorris: done.

lib/shared/bucketing-assembly-script/assembly/types/protobuf/as-generated/DVCUser_PB.ts
jonathannorris: I removed them, as it seemed like generated files shouldn't be part of the repo.

jonathannorris: ahh I think I need to add a brew install script to install `protoc`

rob-odwyer: FWIW, most examples of protobuf I've seen check in the generated files. That might be biased towards Go though, where it's more common to have repos that build with just the base tools right after cloning.
Also means you don't need protoc installed just to build the repo, only if you're changing the protobuf definitions.

jonathannorris: happy to add them back in if you think that makes sense.

PR #428, opened 2023-03-08
chore: add params to provide length for user and variable strings

ajwootto:
- modifications to allow for pre-allocated memory pool approach
- AS will always read the full size of the pre-allocated memory. This results in receiving a string where the first N bytes are valid data and the rest is garbage that is either 0 or leftovers from previously written strings that were longer
- by adding this parameter, we allow the host system to tell WASM how long the "real" string is after it reads in the full set of bytes. We then slice the string to the first N bytes that represent the "actual" string


lib/shared/bucketing-assembly-script/assembly/index.ts
```diff
@@ -37,26 +37,33 @@ const VariableTypeStrings = ['Boolean', 'Number', 'String', 'JSON']
 export function variableForUser(
     sdkKey: string,
     userStr: string,
+    // pass in length of actual underlying string
+    // (the userStr starts with that and may contain extra preallocated bytes)
+    userStrLength: i32,
```
jonathannorris, line 42: Keep the existing `variableForUser` interface and add a new one for this.

jonathannorris (approved):

lib/shared/bucketing-assembly-script/__tests__/bucketing/bucketing.test.ts
```diff
@@ -58,7 +58,9 @@ const variableForUser = (
     { user, variableKey, variableType }:
     { user: unknown, variableKey: string, variableType: VariableType }
 ): SDKVariable | null => {
-    const variableJSON = variableForUser_AS(sdkKey, JSON.stringify(user), variableKey, variableType)
+    const variableJSON = variableForUser_AS(
```
jonathannorris, line 61: We should have our own tests for the new function.

jonathannorris (changes requested):

jonathannorris (approved):

PR #433, opened 2023-03-10
DVC-6132 [fix]: downgrade some JS SDK error logs to warning logs

kaushalkapasi:


sdk/js/src/Client.ts
```diff
@@ -267,7 +267,7 @@ export class DVCClient implements Client {
 
     track(event: ClientEvent): void {
         if (this._closing) {
-            this.logger.error('Client is closing, cannot track new events.')
+            this.logger.warn('Client is closing, cannot track new events.')
```
ajwootto, line 270: this one actually is an error, it's a user problem

sdk/js/src/StreamingConnection.ts
```diff
@@ -17,7 +17,7 @@ export class StreamingConnection {
             this.onMessage(event.data)
         }
         this.connection.onerror = (err) => {
-            this.logger.error('StreamingConnection error', err)
+            this.logger.warn(`StreamingConnection error ${err}`)
```
ajwootto, line 20: maybe instead of saying error, say something like "StreamingConnection interrupted, re-establishing.."

ajwootto: also according to MDN this situation actually means the connection failed to be established in the first place, not that it was closed later. Are we sure this isn't a problem?
https://developer.mozilla.org/en-US/docs/Web/API/EventSource/error_event

sdk/js/src/Client.ts
ajwootto: bump

kaushalkapasi: just fixed! 😂

ajwootto (approved):

PR #434, opened 2023-03-13
DVC-6570 run all bucketing tests against variableForUser() and variableForUserPB()

jonathannorris:


lib/shared/bucketing-assembly-script/__tests__/protobufVariableHelper.ts
```diff
@@ -0,0 +1,129 @@
...
+    _id: string
+    type: number
+    key: string
+    boolValue: boolean
+    doubleValue: number
+    stringValue: string
+}
+
+const pbSDKVariableToJS = (pbSDKVariable: SDKVariable_PB_Type): SDKVariable => {
+    if (pbSDKVariable.type === 0) {
```
chris-hoefgen, line 29: Are you avoiding using the explicit values instead of the VariableType enum here to just keep the tests isolated?

chris-hoefgen (approved):

lib/shared/bucketing-assembly-script/__tests__/protobufVariableHelper.ts
jonathannorris: yea basically I haven't figured out how to properly generate the protobuf Typescript classes yet from the `.proto` file, will do that as part of the NodeJS SDK PR, which will replace all of this.

PR #436, opened 2023-03-14
DVC-6487 switch NodeJS SDK to use protobuf interface for variableForUser()

jonathannorris:
- add `compile-pb` step to the nodeJS build which will generate static classes for our Protobuf types.
- use new `variableForUser_PB()` method in the NodeJS SDK


sdk/nodejs/project.json
```diff
@@ -45,6 +45,24 @@
           }
       ]
     },
+    "compile-pb": {
```
ajwootto, line 48: why does node need its own build step for this? it should be exported from the assembly script lib?

ajwootto (approved):

chris-hoefgen (approved):

PR #438, opened 2023-03-14
Call variableUpdated subscriber with null if variable not in config

emmawillis:
an alternative to https://github.com/DevCycleHQ/js-sdks/pull/430

### DIFF
--- a/sdk/js/src/Client.ts
+++ b/sdk/js/src/Client.ts
@@ -7,6 +7,7 @@ import {
     DVCEvent as ClientEvent,
     DVCUser,
     ErrorCallback,
+    DVCFeature,
 } from './types'
 import { DVCVariable, DVCVariableOptions } from './Variable'
 import { getConfigJson, saveEntity } from './Request'
@@ -25,6 +26,12 @@ import { ConfigRequestConsolidator } from './ConfigRequestConsolidator'
 import { dvcDefaultLogger } from './logger'
 import { DVCLogger } from '@devcycle/types'
 import { StreamingConnection } from './StreamingConnection'
+
+type variableUpdatedHandler = (key: string, variable: DVCVariable<DVCVariableValue> | null) => void
+type featureUpdatedHandler = (key: string, feature: DVCFeature | null) => void
+type newVariablesHandler = () => void
+type errorHandler = (error: unknown) => void
+type initializedHandler = (success: boolean) => void
 
 export class DVCClient implements Client {
     private options: DVCOptions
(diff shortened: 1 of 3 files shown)


sdk/js/src/EventEmitter.ts
```diff
@@ -93,8 +93,9 @@ export class EventEmitter {
                         variable.callback?.call(variable, variable.value)
                     })
                 }
-                this.emit(`${EventNames.VARIABLE_UPDATED}:*`, key, newVariable)
-                this.emit(`${EventNames.VARIABLE_UPDATED}:${key}`, key, newVariable)
+                const finalVariable = newVariable || null
```
ajwootto, line 96: does the event get called if you're never called `client.variable` for a given variable, and it is then removed from the config?

I think that's the bug we need to fix.

emmawillis: yeah it gets called for any variable that was previously in the config but removed in this update, regardless of whether it's used in code anywhere. is that not what we want?

ajwootto: it is, that was just my suspicion of why they were seeing this issue of the event not being triggered at all.

emmawillis: do you think it's better to pass back null in this case or a plain object with what we do know, ie `{key: string, isDefaulted: true}`? the downside to that is sometimes it's called with a DVCVariable instance, and other times a plain object

emmawillis: sounds good

sdk/js/src/EventEmitter.ts
```diff
@@ -93,8 +93,9 @@ export class EventEmitter {
                         variable.callback?.call(variable, variable.value)
                     })
                 }
```
ajwootto, line 95: do you think there's a way to type the subscribe function better, so that when subscribing to the variableUpdated events it would tell you to expect either a `Variable` or `null`?

emmawillis: i've been trying a bunch of things but there's only one subscribe method and the type of `eventHandler` will be different depending on the first argument to subscribe, so think it would have to be something like:

```
type HandlerTypes = {
    VARIABLE_UPDATED: (key: string, variable: DVCVariable<DVCVariableValue> | null) => void,
    ...
}
```
`    subscribe<T extends EventNames>(key: T, handler: Pick<HandlerTypes, T>): void {`
to try to match the string key to the expected type 

but the problem is `EventNames` would have to match the actual regex strings for this to work, ie:
```
type EventNames = 
| `variableUpdated:${string}` 
| `newVariables:${string}` 
| `featureUpdated:${string}` 
| 'initialized' 
| 'error'
```
otherwise it won't accept the regex matching for the key param

... but then i can't index the HandlerTypes type, because i only have these dynamic types like `variableUpdated:${string}`  and not something fixed like VARIABLE_UPDATED

emmawillis: wait never mind i got it working with function overloading

ajwootto: yup overloading is what you want

sdk/js/src/Client.ts
```diff
@@ -26,6 +27,12 @@ import { dvcDefaultLogger } from './logger'
 import { DVCLogger } from '@devcycle/types'
 import { StreamingConnection } from './StreamingConnection'
 
+type variableUpdatedHandler = (key: string, variable: DVCVariable<DVCVariableValue> | null) => void
+type featureUpdatedHandler = (key: string, feature: DVCFeature) => void
```
ajwootto, line 31: what happens here when you get removed from a feature?

emmawillis: good point, it's undefined too - switched it to null

ajwootto (approved):

PR #440, opened 2023-03-15
DVC-6132 [fix]: update realtime connection onerror logs to be warning

kaushalkapasi:
- update realtime connection onerror logs to always be warning logs
- update config fetch error / error log to capture the response errors

### DIFF
--- a/sdk/js/src/Request.ts
+++ b/sdk/js/src/Request.ts
@@ -92,9 +92,10 @@ export const getConfigJson = async (
         const res = await get(url)
         return res.data
     } catch (ex: any) {
+        const errorString = JSON.stringify(ex?.response?.data.data.errors)
         logger.error(`Request to get config failed for url: ${url}, ` +
-            `response message: ${ex.message}, response data: ${ex?.response?.data}`)
-        throw new Error('Failed to download DevCycle config.')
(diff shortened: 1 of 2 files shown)


sdk/js/src/Request.ts
```diff
@@ -93,8 +93,9 @@ export const getConfigJson = async (
         return res.data
     } catch (ex: any) {
         logger.error(`Request to get config failed for url: ${url}, ` +
-            `response message: ${ex.message}, response data: ${ex?.response?.data}`)
-        throw new Error('Failed to download DevCycle config.')
+            `response message: ${ex.message}, response data: ${JSON.stringify(ex?.response?.data.data.errors)}`)
+        throw new Error(`Failed to download DevCycle config. ` +
+            `Error details: ${JSON.stringify(ex?.response?.data.data.errors)}`)
```
laurawarr, line 98: just a nitpick, but you could store this to a variable so that we don't need to duplicate stringifying the errors

laurawarr (approved):

sdk/js/src/Request.ts
kaushalkapasi: will update, makes sense :)

laurawarr (approved):

PR #442, opened 2023-03-15
Add configUpdated event

emmawillis:


sdk/js/src/EventEmitter.ts
```diff
@@ -23,7 +24,7 @@ export class EventEmitter {
         checkParamType('key', key, 'string')
         checkParamType('handler', handler, 'function')
 
-        const eventNames = Object.keys(EventNames).map((e) => e.toLowerCase())
+        const eventNames = Object.values(EventNames)
         if (!eventNames.includes(key) &&
```
emmawillis, line 28: I'm pretty sure this was a bug - using the EventNames keys instead of values meant that, for ex, `variable_updated` was a valid event to subscribe to, but it never receives any events

sdk/js/src/EventEmitter.ts
```diff
@@ -39,7 +40,7 @@ export class EventEmitter {
     unsubscribe(key: string, handler?: eventHandler): void {
         checkParamType('key', key, 'string')
 
-        const eventNames = Object.keys(EventNames).map((e) => e.toLowerCase())
+        const eventNames = Object.values(EventNames)
         if (!eventNames.includes(key)) {
```
ajwootto, line 44: kinda weird that the check here is different than the one on subscribe

ajwootto: doesnt this mean you cant unsub from "variableUpdated:*"?

emmawillis: good catch, i added made the check the same in both

ajwootto (approved):

PR #446, opened 2023-03-17
Dispatch real events in eventEmitter

emmawillis:
this means the handler is called in a different thread than our regular config fetching code, so we don't catch exceptions thrown in the callback. 

every time subscribe/ unsubscribe are called, i had to remove the whole event listener, update our event handler store, and then add the listener again - not sure if there's a better way to do this? but since i'm using a wrapper to call multiple handlers, there doesn't seem to be a better way to edit the list of handlers in place

### DIFF
--- a/sdk/js/src/EventEmitter.ts
+++ b/sdk/js/src/EventEmitter.ts
@@ -20,10 +20,20 @@ const isInvalidEventKey = (key: string): boolean => {
     !key.startsWith(EventNames.NEW_VARIABLES)
 }
 
-export class EventEmitter {
-    events: Record<string, eventHandler[]>
+
+const callHandler = (event: Event, handler: eventHandler): void  => {
+    const args = (<CustomEvent<any[]>>event).detail
+    handler(...args)
+}
+
+export class EventEmitter extends EventTarget {
+    events: Record<string, {
+        handler: eventHandler,
+        handlerWrapper: eventHandler
+    }[]>
 
     constructor() {
+        super()
         this.events = {}
     }
 
@@ -33,10 +43,14 @@ export class EventEmitter {
 
         if (isInvalidEventKey(key)) {
             throw new Error('Not a valid event to subscribe to')
-        } else if (!this.events[key]) {
-            this.events[key] = [ handler ]
+        }  
+        
+        const handlerWrapper = (event: Event) => callHandler(event, handler)
+        this.addEventListener(key, handlerWrapper)
+        if (!this.events[key]) {
+            this.events[key] = [ { handler, handlerWrapper } ]
         } else {
-            this.events[key].push(handler)
+            this.events[key].push({ handler, handlerWrapper })
         }
     }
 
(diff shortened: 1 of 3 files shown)


sdk/js/src/EventEmitter.ts
```diff
@@ -20,10 +20,19 @@ const isInvalidEventKey = (key: string): boolean => {
...
-export class EventEmitter {
+const callHandlers = (event: Event, handlers: eventHandler[]): void  => {
+    const args = (<CustomEvent<any[]>>event).detail
+
+    handlers.forEach((handler) => {
+        handler(...args)
+    })
+}
+
+export class EventEmitter extends EventTarget {
```
ajwootto, line 31: are there any potential issues with browser compatibility that we introduce by adding this? it's a pretty new thing

sdk/js/src/EventEmitter.ts
```diff
@@ -33,37 +42,37 @@ export class EventEmitter {
 
         if (isInvalidEventKey(key)) {
             throw new Error('Not a valid event to subscribe to')
-        } else if (!this.events[key]) {
+        }  
+
+        this.removeEventListener(key, (event) => callHandlers(event, this.events[key]))
```
ajwootto, line 47: might wanna make sure that `subscribe` is bound to the `this` context, otherwise this line would break if someone did something like 

```
const { subscribe } = client
subscribe('event', () => {})
```

ajwootto: actually nvm this is the internal method I guess you can't do that

sdk/js/src/EventEmitter.ts
emmawillis: based on this https://developer.mozilla.org/en-US/docs/Web/API/EventTarget#browser_compatibility all major browsers have supported the EventTarget class since at least 2020, do you think that's good enough?

emmawillis: oh wait but vercel can't build it for some reason, `(node:1454) UnhandledPromiseRejectionWarning: ReferenceError: EventTarget is not defined`
i guess because we build it with node? maybe i need to import it with yarn somehow for node

emmawillis: oh no i actually think it's because vercel's version of node is 14.x and we need 14.5 or higher

ajwootto: shouldn't this be running in a browser context? How come node is involved?

emmawillis: yeah i'm confused about that too - i guess we build it with node? it's published to npm right?

ajwootto: it is, but the build should succeed regardless. This looks more like a runtime exception from actually running it (inside Node)

emmawillis: hmm i switched vercel to use node 16 for this project and it built it no problem

suthar26 (approved):

PR #459, opened 2023-03-24
Update publishing instructions in readme

emmawillis:


README.md
```diff
@@ -99,15 +99,22 @@ You can run commands in every javascript project using Nx: `nx run-many --target
 This allows you to lint and/or test all projects at once.
 
 ### Publishing a Release
-To publish a release, use lerna to create new versions of all changed packages (ensure you do this on the main branch)
+Setup:
+- run `aws sso login`
+- run `./scripts/brew-install.sh`
+- ensure you're on the main branch with the latest code 
+
```
aurorawalker, line 106: The addition of this section is great, but I think it actually needs to live higher up in the generic 'Setup' section rather than down here for publishing.

It's very similar to the setup in the DevCycle API repo (https://github.com/DevCycleHQ/DevCycle#setup) and also depends on having already done `aws configure sso` (which you likely had already done!)

My suggestion is to move this setup content higher, and maybe can do a link to the DevCycle repo that says something along the lines of "Make sure you have already done `aws configure sso` like in the other project... and then continue with the instructions.

aurorawalker: Edit - I would still move that section up, but can make it more generic (i.e. not link to a private repo) by simply stating that the first step for setup is `aws configure sso` then `aws sso login`

README.md
```diff
@@ -99,15 +99,22 @@ You can run commands in every javascript project using Nx: `nx run-many --target
...
+- run `./scripts/brew-install.sh`
+- ensure you're on the main branch with the latest code 
+ 
+Then use lerna to create new versions of all changed packages (ensure you do this on the main branch)
 `yarn lerna:version`
 
-Push up the new tags and version changes, then run:
+This will automatically push the latest tags and version updates to github. 
 
+To publish the versions to npm, you need a one-time password from our NPM account. Search for 'npm' in 1Password and copy the OTP under `Npmjs (taplyticsuser)`
```
aurorawalker, line 112: I think moving this line up here is a great improvement (nothing like instructions that say "run a command, but before you do that..." after you run the command :P ) but can drop the reference to 1pass

README.md
emmawillis: i think `aws configure sso` makes sense to be at the top, but i want to keep the login part in the release section because part of what tripped me up was that i had been logged out at the time i tried to publish (and the error message for that was super unclear)

PR #464, opened 2023-03-28
Remove JSON library from variable evaluation path.

ajwootto:
- removes the double allocation on variable evaluations by directly evaluating against the Protobuf classes instead of converting to JSON obj
- fix the downstream effects of doing that:
  - setting client custom data now requires protobuf instead of JSON
  - generateBucketedConfig now requires protobuf instead of JSON
  - client custom data is stored in both Protobuf and JSON form (JSON form is used for events)
- fix tests to use protobuf instead of json


lib/shared/bucketing-assembly-script/assembly/tsconfig.json
```diff
@@ -2,5 +2,8 @@
   "extends": "../../../../node_modules/assemblyscript/std/assembly.json",
   "include": [
     "./**/*.ts"
-  ]
+  ],
+  "compilerOptions": {
+    "strictPropertyInitialization": false
```
ajwootto, line 7: this gets rid of errors from defining classes that only have properties and no constructor, which is valid Assemblyscript but was getting type errors..

lib/shared/bucketing-assembly-script/assembly/types/bucketedUserConfig.ts
```diff
@@ -220,10 +220,14 @@ export class SDKVariable extends JSON.Obj {
         const stringValue = (this.type === 'String' && this.value.isString)
             ? (this.value as JSON.Str).valueOf()
             : null
-        const jsonValue = (this.type === 'JSON' && this.value.isObj)
+        let jsonValue = (this.type === 'JSON' && this.value.isObj)
             ? this.value.stringify()
             : null
 
+        if (this.type === 'JSON' && jsonValue == null && this.value.isString) {
```
ajwootto, line 227: I think this was actually a bug. It seems like this was just setting the value to null for all json variables, since the value field is always a string in this context

lib/shared/bucketing-assembly-script/assembly/types/dvcUserPB.ts
```diff
@@ -0,0 +1,100 @@
...
+    readonly platformVersion: string
+    readonly sdkType: string
+    readonly sdkVersion: string
+    readonly hostname: string | null
+
+    constructor(user: DVCUser_PB) {
+        this.user_id = user.userId
+        this.email = deNullString(user.email)
+        this.name = deNullString(user.name)
+        this.language = deNullString(user.language)
```
ajwootto, line 60: would this function call be pass by reference or by value for these strings? I wonder if calling a helper here instead of doing it inline actually results in an extra allocation?

lib/shared/bucketing-assembly-script/assembly/types/dvcUserPB.ts
```diff
@@ -0,0 +1,100 @@
+import { DVCUser_PB } from './protobuf-generated/DVCUser_PB'
+import { NullableString } from './protobuf-generated/NullableString'
+import { NullableDouble } from './protobuf-generated/NullableDouble'
+import { CustomDataValue } from './protobuf-generated/CustomDataValue'
+import { _getPlatformData } from '../managers/platformDataManager'
+import { NullableCustomData } from './protobuf-generated/NullableCustomData'
+import { CustomDataType } from './protobuf-generated/CustomDataType'
+
+export class CustomDataValueInterpreter {
```
ajwootto, line 9: this basically mimics the methods we had on the JSON CustomData class. They are static because it seems i can't extend the CustomDataValue PB class to add more methods to it and then cast one to the other...I'd have to copy the PB class contents to a new instance of my extended class. That results in a bunch of extra allocations we don't need.

lib/shared/bucketing-assembly-script/assembly/managers/clientCustomDataManager.ts
```diff
@@ -1,16 +1,27 @@
+import { CustomDataValue } from '../types'
 import { JSON } from 'assemblyscript-json/assembly'
+import { getJSONObjFromCustomDataMap } from '../helpers/jsonHelpers'
 
-const _clientCustomData: Map<string, JSON.Obj> = new Map()
+const _clientCustomData: Map<string, Map<string, CustomDataValue>> = new Map()
```
ajwootto, line 5: I'm storing the client custom data in both Map and JSON form. See the comment in eventQueueManager below for an explanation.

lib/shared/bucketing-assembly-script/assembly/helpers/jsonHelpers.ts
```diff
@@ -210,3 +212,20 @@ export function isFlatJSONObj(json: JSON.Obj | null): bool {
     }
     return true
 }
+
+export function getJSONObjFromCustomDataMap(customData: Map<string, CustomDataValue>): JSON.Obj {
```
ajwootto, line 216: pretty straightforward converter to go from the Map of CustomDataValue that we get from decoding Protobuf, to a JSON Obj

lib/shared/bucketing-assembly-script/assembly/bucketing/segmentation.ts
```diff
@@ -4,14 +4,14 @@ import { OptionsType, versionCompare } from './versionCompare'
...
-    DVCPopulatedUser,
     validSubTypes,
     CustomDataFilter,
     UserFilter,
-    NoIdAudience, AudienceMatchFilter
+    NoIdAudience, AudienceMatchFilter, CustomDataValue
 } from '../types'
 import { JSON } from 'assemblyscript-json/assembly'
 import { getF64FromJSONValue } from '../helpers/jsonHelpers'
+import { CustomDataValueInterpreter, DVCPopulatedUserPB } from '../types/dvcUserPB'
```
jonathannorris, line 14: you should just export these types in `types/index.ts`

lib/shared/bucketing-assembly-script/protobuf/variableForUserParams.proto
```diff
@@ -39,6 +39,10 @@ message NullableCustomData {
   bool isNull = 2;
 }
 
+message ClientCustomData_PB {
```
jonathannorris, line 42: where is this used?

ajwootto: for passing data into the setClientCustomData method (which now uses protobuf instead of JSON)

lib/shared/bucketing-assembly-script/assembly/bucketing/segmentation.ts
```diff
@@ -406,16 +416,20 @@ export function getFilterValues(filter: UserFilter): JSON.Value[] {
...
-        && (!stringValue || stringValue.valueOf() !== '')
-        && (!floatValue || !isNaN(floatValue.valueOf()))
-        && (!intValue || !isNaN(intValue.valueOf()))
+    if (CustomDataValueInterpreter.isString(value)) {
+        const stringValue = CustomDataValueInterpreter.asString(value)
+        return stringValue !== null && stringValue !== ''
+    }
+    if (CustomDataValueInterpreter.isFloat(value)) {
+        const floatValue = CustomDataValueInterpreter.asNumber(value)
+        return floatValue !== null && !isNaN(floatValue)
```
jonathannorris, line 427: I guess this means there are no more int values for customData values, I wonder how that affects things...

lib/shared/bucketing-assembly-script/assembly/bucketing/segmentation.ts
```diff
@@ -406,16 +416,20 @@ export function getFilterValues(filter: UserFilter): JSON.Value[] {
...
-    return value !== null
-        && !!(stringValue || floatValue || intValue || boolValue)
-        && (!stringValue || stringValue.valueOf() !== '')
-        && (!floatValue || !isNaN(floatValue.valueOf()))
-        && (!intValue || !isNaN(intValue.valueOf()))
+    if (CustomDataValueInterpreter.isString(value)) {
+        const stringValue = CustomDataValueInterpreter.asString(value)
+        return stringValue !== null && stringValue !== ''
+    }
+    if (CustomDataValueInterpreter.isFloat(value)) {
```
jonathannorris, line 425: should these all be `else if` statments?

lib/shared/bucketing-assembly-script/assembly/bucketing/segmentation.ts
ajwootto: I don't think there were anyway, since we were using protobuf to construct the user JSON object and protobuf didnt have ints

lib/shared/bucketing-assembly-script/assembly/types/dvcUserPB.ts
ajwootto: yeah it is I checked the WAT file

lib/shared/bucketing-assembly-script/assembly/testHelpers.ts
```diff
@@ -64,23 +68,27 @@ export function checkVersionFiltersFromJSON(appVersion: string | null, filterStr
     return _checkVersionFilters(appVersion, filter)
 }
 
-export function checkCustomDataFromJSON(data: string | null, filterStr: string): bool {
+export function checkCustomDataFromJSON(data: Uint8Array | null, filterStr: string): bool {
```
jonathannorris, line 71: just remove `FromJSON`

lib/shared/bucketing-assembly-script/assembly/testHelpers.ts
```diff
@@ -64,23 +68,27 @@ export function checkVersionFiltersFromJSON(appVersion: string | null, filterStr
...
-        throw new Error('checkCustomDataFromJSON data param not a JSON Object')
-    }
 
     const filter = new CustomDataFilter(filterJSON as JSON.Obj)
-    const dataJSONObj = dataJSON && dataJSON.isObj ? dataJSON as JSON.Obj : null
-    return _checkCustomData(dataJSONObj, new JSON.Obj(), filter)
+    return _checkCustomData(user, new Map<string, CustomDataValue>(), filter)
 }
 
 export function evaluateOperatorFromJSON(
```
jonathannorris, line 89: `evaluateOperator`

lib/shared/bucketing-assembly-script/assembly/testHelpers.ts
```diff
@@ -64,23 +68,27 @@ export function checkVersionFiltersFromJSON(appVersion: string | null, filterStr
...
     operatorStr: string,
```
jonathannorris, line 90: `operatorJSONStr`

lib/shared/bucketing-assembly-script/assembly/types/bucketedUserConfig.ts
jonathannorris:
```suggestion
        if (jsonValue == null && this.type === 'JSON' && this.value.isString) {
```

lib/shared/bucketing-assembly-script/assembly/types/dvcUser.ts
```diff
@@ -257,18 +237,14 @@ export class DVCPopulatedUser extends JSON.Value implements DVCUserInterface {
         return new DVCPopulatedUser(DVCUser.fromUTF8(userBytes))
     }
 
-    getCombinedCustomData(): JSON.Obj {
-        return this._combinedCustomData
-    }
```
jonathannorris, line 262: Can we keep the custom data logic in this class so that we don't mess up the order of checking customData and customData

jonathannorris:
* main
  * **PR #464** <a href="https://app.graphite.dev/github/pr/DevCycleHQ/js-sdks/464" target="_blank">Graphite</a>  👈
    * **PR #471** <a href="https://app.graphite.dev/github/pr/DevCycleHQ/js-sdks/471" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/DevCycleHQ/js-sdks/464?utm_source=stack-comment).

lib/shared/bucketing-assembly-script/assembly/managers/eventQueueManager.ts
```diff
@@ -102,13 +103,15 @@ export function onPayloadFailure(sdkKey: string, payloadId: string, retryable: b
     requestPayloadManager.markPayloadFailure(payloadId, retryable)
 }
 
-export function queueEvent(sdkKey: string, userStr: string, eventStr: string): void {
+export function queueEvent(sdkKey: string, userStr: Uint8Array, eventStr: string): void {
     const eventQueue = getEventQueue(sdkKey)
-    const dvcUser = DVCPopulatedUser.fromJSONString(userStr)
+    const data = decodeDVCUser_PB(userStr)
+    const dvcUser = new DVCPopulatedUserPB(data)
```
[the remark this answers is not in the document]
jonathannorris: added a code comment.

jonathannorris (approved):

PR #472, opened 2023-04-04
DVC-6846 [fix]: add disableRealtimeUpdates options to JS SDK

kaushalkapasi:


sdk/js/src/Client.ts
```diff
@@ -423,11 +423,15 @@ export class DVCClient implements Client {
...
-                        this.logger
-                    )
+                    if (!this.options.disableRealtimeUpdates) {
+                        this.streamingConnection = new StreamingConnection(
+                            this.config.sse.url,
+                            this.onSSEMessage.bind(this),
+                            this.logger
+                        )
+                    } else {
+                        this.logger.info('Realtime updates are disabled')
```
jonathannorris, line 433: should this be a debug log? probably not something ppl want logging every time they launch the SDK with info logs.

kaushalkapasi: Since this is an active choice that changes the SDKs functionality, decided to use an `info` level log so it doesn't get lost in debug level logging

jonathannorris: Maybe update the wording of the log to say its disabled because of the start option? so it doesn't look like an error.

jonathannorris (approved):

PR #475, opened 2023-04-14
DVC-6993 update JS SDKs to send sseEtag value

jonathannorris:


sdk/js/src/Request.ts
```diff
@@ -79,12 +79,13 @@ export const getConfigJson = async (
...
     logger: DVCLogger,
     options?: DVCOptions,
-    extraParams?: {sse: boolean, lastModified?: number}
+    extraParams?: { sse: boolean, lastModified?: number, etag?: string }
 ): Promise<BucketedUserConfig> => {
     const edgeDBParam = options?.enableEdgeDB ? ('&enableEdgeDB=' + options.enableEdgeDB): ''
     const sseParam = extraParams?.sse ? '&sse=1' : ''
     const lastModified = extraParams?.lastModified ? `&sseLastModified=${extraParams.lastModified}` : ''
-    const queryParams = `${serializeUser(user)}${edgeDBParam}${sseParam}${lastModified}`
+    const etag = extraParams?.etag ? `&sseEtag=${extraParams.etag}` : ''
```
ajwootto, line 87: any chance this ends up not being URL safe?

jonathannorris: I don't think so, maybe with the `W/` ones? https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag

ajwootto: yeah that would probably break it.. you should probably URL encode this just to be safe

sdk/js/src/utils.ts
```diff
@@ -13,15 +13,13 @@ const convertToQueryFriendlyFormat = (property?: any): any => {
...
-        .reduce((prev: string, curr: string, _: number): string => {
-            const userProperty = convertToQueryFriendlyFormat(user[curr as keyof DVCClientAPIUser])
-            const nextQueryParam = userProperty !== null && userProperty !== undefined
-                ? `${prev && '&'}${curr}=${encodeURIComponent(userProperty)}`
-                : ''
-            return `${prev}${nextQueryParam}`
-        }, '')
+export const serializeUserSearchParams = (user: DVCClientAPIUser, queryParams: URLSearchParams): void => {
+    Object.keys(user).forEach((key) => {
+        const userProperty = convertToQueryFriendlyFormat(user[key as keyof DVCClientAPIUser])
```
ajwootto, line 18: this keyof stuff is a good reason to use `for const key in user` instead since the types will be inferred

ajwootto (approved):

PR #481, opened 2023-04-25
chore: adopt existing signature, add types

nikolasleblanc:
Noticed that variableUpdated was passing the key and the variable object to the handler so did the same with variableEvaluated, also updated the subscribe overrides and the accompanying jsdocs.

### DIFF
--- a/sdk/js/src/Client.ts
+++ b/sdk/js/src/Client.ts
@@ -29,7 +29,7 @@ import { StreamingConnection } from './StreamingConnection'
 
 type variableUpdatedHandler = (
     key: string,
-    variable: DVCVariable<DVCVariableValue>
+    variable: DVCVariable<DVCVariableValue> | null
 ) => void
 type featureUpdatedHandler = (key: string, feature: DVCFeature | null) => void
 type newVariablesHandler = () => void
--- a/sdk/js/src/Client.ts
+++ b/sdk/js/src/Client.ts
@@ -36,6 +36,10 @@ type newVariablesHandler = () => void
 type errorHandler = (error: unknown) => void
 type initializedHandler = (success: boolean) => void
 type configUpdatedHandler = (newVars: DVCVariableSet) => void
+type variableEvaluatedHandler = (
+    key: string,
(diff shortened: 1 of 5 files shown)


ajwootto (approved):

sdk/js/src/Client.ts
```diff
@@ -36,6 +36,10 @@ type newVariablesHandler = () => void
 type errorHandler = (error: unknown) => void
 type initializedHandler = (success: boolean) => void
 type configUpdatedHandler = (newVars: DVCVariableSet) => void
+type variableEvaluatedHandler = (
+    key: string,
+    variable: DVCVariable<DVCVariableValue> | null
```
ajwootto, line 41: this should never be null

ajwootto (changes requested):

sdk/js/src/Client.ts
```diff
@@ -29,13 +29,17 @@ import { StreamingConnection } from './StreamingConnection'
 
 type variableUpdatedHandler = (
     key: string,
-    variable: DVCVariable<DVCVariableValue> | null
+    variable: DVCVariable<DVCVariableValue>
```
ajwootto, line 32: this can be null

sdk/js/src/types.ts
```diff
@@ -201,6 +196,8 @@ export interface DVCClient {
      *  - `variableUpdated:<variable.key>` -> (key: string, variable: DVCVariable)
      *  - `featureUpdated:*` -> (key: string, feature: DVCFeature)
      *  - `featureUpdated:<feature.key>` -> (key: string, feature: DVCFeature)
+     *  - `variableEvaluated:*` -> (key: string, variable: DVCVariable)
+     *  - `variableEvaluated:<varable.key>` -> (key: string, variable: DVCVariable)
```
kaushalkapasi, line 200: typo:
```suggestion
     *  - `variableEvaluated:<variable.key>` -> (key: string, variable: DVCVariable)
```

nikolasleblanc: Fixed!

kaushalkapasi (approved):

PR #487, opened 2023-05-03
add arrayvariablevalue type

elliotCamblor:
This is done so that the config can contain array values for variables.

### DIFF
--- a/lib/shared/types/src/types/config/models/variable.ts
+++ b/lib/shared/types/src/types/config/models/variable.ts
@@ -40,7 +40,8 @@ export enum VariableType {
  * Supported variable values
  */
 export type DVCJSON = { [key: string]: string | boolean | number }
-export type VariableValue = string | boolean | number | DVCJSON
+export type VariableValue = string | boolean | number | Record<string, unknown> |
+    Array<string | boolean | number | Record<string, unknown>>
 
 type UnionToIntersection<U> =
     (U extends any ? (k: U)=>void : never) extends ((k: infer I)=>void) ? I : never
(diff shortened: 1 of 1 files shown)


lib/shared/types/src/types/config/models/variable.ts
```diff
@@ -41,6 +41,7 @@ export enum VariableType {
  */
 export type DVCJSON = { [key: string]: string | boolean | number }
 export type VariableValue = string | boolean | number | DVCJSON
+export type ArrayVariableValue = VariableValue | Array<string | boolean | number>
```
ajwootto, line 44: arrays can contain any valid JSON so we should make sure DVCJSON is in here

ajwootto: actually one thing I just realized is that the VariableValue type definition shouldn't even include DVCJSON, since we don't restrict variable values to one level of nesting. I think the correct type there would just be a `Record<string, unknown>` or something

lib/shared/types/src/types/config/models/variable.ts
```diff
@@ -40,7 +40,8 @@ export enum VariableType {
  * Supported variable values
  */
 export type DVCJSON = { [key: string]: string | boolean | number }
-export type VariableValue = string | boolean | number | DVCJSON
+export type VariableValue = string | boolean | number | Record<string, unknown> |
```
ajwootto, line 43: are these used by the JS or Node SDKs at all? If so I'd probably separate the type for those so that it's not possible to use arrays on those SDKs as default values

elliotCamblor: I'm really not sure how thats going to work without completely overhauling all of the types. Ultimately the defaultValue type has to be same as the variable value type, and I dont see how I can decouple those from the variable value types in the bucketing config

PR #496, opened 2023-06-07
DVC-4148 deferred mode

ajwootto:
- add "deferred mode" for React and JS which allows the client to be instantiated without immediately sending a config call or requiring user data to be ready
- the client is then "initialized" later when the first call to "identifyUser" is made
- this allows frontend applications to separate client creation from user identification so that the client can be passed down through the Provider context without having a user yet. Any calls to variable will just return default values until the user is identified. Calls to `useIsDVCInitialized` will correctly return "false" until the user is identified for the first time.


sdk/js/src/Client.ts
```diff
@@ -138,6 +137,84 @@ export class DVCClient<
...
+            this.options,
+            undefined,
+            storedAnonymousId,
+        )
+
+        if (!this.options.disableConfigCache) {
+            await this.getConfigCache(this.user)
+        } else {
+            this.logger.info('Skipping config cache')
+        }
```
jonathannorris, line 165: we should probably put the log for `this.options.disableConfigCache` in `getConfigCache()`

jonathannorris (approved):

elliotCamblor (approved):

jsalaber (approved):

PR #501, opened 2023-06-08
fix: DVC-7694 update OpenFeature provider to support 1.3.0

jonathannorris:
- Update OpenFeature provider to support changes in spec: https://github.com/open-feature/js-sdk/releases/tag/js-sdk-v1.3.0
- adds support for `initialize()` and `onClose()` methods
- adds support for `flagMetadata` setting params: `type` / `isDefaulted` / `defaultValue`

### DIFF
--- a/sdk/openfeature-nodejs-provider/src/DevCycleProvider.ts
+++ b/sdk/openfeature-nodejs-provider/src/DevCycleProvider.ts
@@ -180,9 +180,10 @@ export default class DevCycleProvider implements Provider {
     ): ResolutionDetails<T> {
         const flagMetadata: FlagMetadata = {
             isDefaulted: variable.isDefaulted,
-            defaultValue: (typeof variable.defaultValue === 'object')
-                ? JSON.stringify(variable.defaultValue)
-                : variable.defaultValue
+            defaultValue:
+                typeof variable.defaultValue === 'object'
+                    ? JSON.stringify(variable.defaultValue)
+                    : variable.defaultValue,
         }
         if (variable.type) {
             flagMetadata.type = variable.type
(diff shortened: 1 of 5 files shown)


sdk/openfeature-nodejs-provider/src/DevCycleProvider.ts
```diff
@@ -165,11 +178,22 @@ export default class DevCycleProvider implements Provider {
     private resultFromDVCVariable<T>(
         variable: DVCVariable,
     ): ResolutionDetails<T> {
+        const flagMetadata: FlagMetadata = {
```
jonathannorris, line 181: @toddbaert / @beeme1mr is this the kind of data you were expecting to be included in `flagMetadata`?

beeme1mr: The data returned in `flagMetadata` was intended to be used for debugging and telemetry. I'm not sure how valuable returning information about defaults is but it's completely up to you. Perhaps instead of returning `isDefaulted` you could set the [resolution detail](https://openfeature.dev/specification/types) to `DEFAULT`.

I'm not sure if you have access to this information in the SDK but including things like the organization, project, and environment would be very useful.

chris-hoefgen (approved):

sdk/openfeature-nodejs-provider/src/DevCycleProvider.ts
jonathannorris: We already set the resolution detail, I guess these aren't needed then. We do have access to org / project / env. however, those aren't publically in our SDK, we will need to think about how to do that.

PR #502, opened 2023-06-08
chore: update to latest nx

ajwootto:
- revert upgrade to Yarn for now because nx migration failed with newer yarn version
https://github.com/nrwl/nx/issues/16338
- run nx migrations
- remove attempt at ESM support from types package for now
- update tsconfig to use `node16` module and resolution by default (strictest) while using `esnext` for frontend-only packages (react & react native)


lib/shared/bucketing-test-data/src/main.ts
```diff
@@ -3,20 +3,24 @@ import path from 'path'
 import * as testData from './data/testData'
 import * as largeTestData from './data/largeConfig'
 
-const folderName = '../../../../../lib/shared/bucketing-test-data/json-data/'
-const folderPath = path.resolve(__dirname, folderName)
+const folderName = 'lib/shared/bucketing-test-data/json-data/'
+const folderPath = path.resolve(process.env.NX_WORKSPACE_ROOT || '', folderName)
+console.log(process.env.NX_WORKSPACE_ROOT)
```
kaushalkapasi, line 8: console.logs 👀

kaushalkapasi:
Overall, looks good to me... just need to delete the `Vercel - js-sdks-async-with-provider` task from the GH file so it doesn't try to deploy a project that doesn't exist 👀

kaushalkapasi (approved):

PR #508, opened 2023-06-16
feat(shared-bucketing-as): add echoString and echoUint8Array to bucketing-assembly-script

rob-odwyer:
Planning to use these to test the methods in the Python implementation for reading and writing strings and byte arrays.

### DIFF
--- a/lib/shared/bucketing-assembly-script/assembly/testHelpers.ts
+++ b/lib/shared/bucketing-assembly-script/assembly/testHelpers.ts
@@ -148,6 +148,14 @@ export function testBucketedUserConfigClass(userConfigStr: string): string {
     return userConfig.stringify()
 }
 
+export function echoString(str: string): string {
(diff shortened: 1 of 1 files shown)


lib/shared/bucketing-assembly-script/assembly/testHelpers.ts
```diff
@@ -148,6 +148,18 @@ export function testBucketedUserConfigClass(userConfigStr: string): string {
     return userConfig.stringify()
 }
 
+export function echoString(str: string): string {
+    return str
+}
+
+export function echoUint8Array(str: Uint8Array): Uint8Array {
```
chris-hoefgen, line 155: minor nitpick maybe call the variable something other than 'str' here since you use that above.  Maybe something else like 'data'?

chris-hoefgen (approved):

PR #513, opened 2023-06-26
fix: type generic arg on js client

ajwootto:
- fix not being able to pass a type generic to the `initialize` call for the JS SDK
- fix the definition of DVCJSON to accurately reflect what the platform currently supports.

### DIFF
--- a/lib/shared/types/src/types/config/models/variable.ts
+++ b/lib/shared/types/src/types/config/models/variable.ts
@@ -39,7 +39,9 @@ export enum VariableType {
 /**
  * Supported variable values
  */
-export type DVCJSON = { [key: string]: string | boolean | number }
+export type DVCJSON = {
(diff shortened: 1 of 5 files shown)


lib/shared/types/src/types/config/models/variable.ts
```diff
@@ -39,7 +39,9 @@ export enum VariableType {
 /**
  * Supported variable values
  */
-export type DVCJSON = { [key: string]: string | boolean | number }
+export type DVCJSON = {
+    [key: string]: string | boolean | number | DVCJSON | unknown[] | null
```
jonathannorris, line 43: What does `unknown[] | null` here functionally do? Does this mean we let folks set null / undefined as default values?

jonathannorris (approved):

PR #526, opened 2023-07-10
feat: Create release action to handle automated releases.

JamieSinn:
WIP - Working on getting the final workflow setup and whether or not we want to specify a specific tag to override yet.


.prettierignore
```diff
@@ -8,7 +8,10 @@ jest.preset.js
 *.MD
 lib/shared/bucketing-assembly-script/
 .yarn
-.husky
+.huskyx
```
jonathannorris, line 11: I think this is a typo

.prettierignore
```diff
@@ -8,7 +8,10 @@ jest.preset.js
...
 .verdaccio
 node_modules
 **/.next/**
+.github/**
+nx.json
+lerna.json
```
jonathannorris, line 17: why do these need to be ignored?

nx.json
```diff
@@ -1,108 +1,134 @@
...
-                    "build",
-                    "test",
-                    "lint",
-                    "check-types",
-                    "package",
-                    "prepare",
-                    "build:json",
-                    "compile-pb-js",
-                    "compile-pb-as"
-                ],
```
jonathannorris, line 16: why did all of these get removed?

scripts/npm-safe-publish.sh
```diff
@@ -68,36 +66,32 @@ if [[ "$NPM_SHOW" != "$NPM_LS" ]]; then
...
-      -d api_key=$DEVCYCLE_PROD_SLEUTH_API_TOKEN \
-      -d environment=production \
-      -d sha=$SHA https://app.sleuth.io/api/1/deployments/taplytics/js-sdks-2/register_deploy
   else
     echo "NPM Publish Local"
     npm publish --local
   fi
 else
   echo "Versions are the same ($NPM_SHOW = $NPM_LS). Not pushing"
+  exit 0
```
jonathannorris, line 94: I think we want it to go down the Deprecated code path even if the existing versions are the same.

nx.json
```diff
@@ -1,108 +1,134 @@
...
-                "parallel": 1
-            }
-        }
+  "tasksRunnerOptions": {
```
jonathannorris, line 2: There are a couple of changes in this file? Were they intentional or a messed up re-base?

.prettierignore
JamieSinn: Yup

nikolasleblanc:
@CodiumAI-Agent /review

nikolasleblanc:
@CodiumAI-Agent /improve

.prettierignore
JamieSinn: They're being errored against - even though they're correctly formatted.

nx.json
JamieSinn: Not sure - Lerna/NX did something to integrate the two - not sure why or what it did

scripts/npm-safe-publish.sh
JamieSinn: If the existing versions are the same - nothing should be published - NPM Publish will throw an exit code if we do - which is why I wanted to terminate early.

nx.json
JamieSinn: Lerna+NX did some integration stuff that I'm not sure what exactly was done

PR #529, opened 2023-07-18
fix: [DVC-8213] various fixes from JS SDK release, update example apps

jonathannorris:


sdk/nodejs/src/index.ts
```diff
@@ -12,6 +12,10 @@ export { dvcDefaultLogger } from './utils/logger'
  * @deprecated Use DevCycleClient instead
  */
 export type DVCClient = DevCycleClient
+/**
+ * @deprecated Use DevCycleCloudClient instead
+ */
+export type DVCCloudClient = DevCycleCloudClient
```
jonathannorris, line 18: this old type was missed in the original deploy

scripts/npm-safe-publish.sh
```diff
@@ -104,23 +104,34 @@ fi
 if [[ "$DEPRECATED_PACKAGE" != "" ]]; then
   echo "Deploy to Deprecated Package: $DEPRECATED_PACKAGE"
 
-  # Backup the original package.json
-  cp package.json package.json.bak
+  # Check version for DEPRECATED_PACKAGE
```
jonathannorris, line 107: @JamieSinn checkout these script updates.

scripts/npm-safe-publish.sh
```diff
@@ -104,23 +104,34 @@ fi
...
+  NPM_SHOW_DEPRECATED="$(npm show "$DEPRECATED_PACKAGE" version)"
```
JamieSinn, line 108: Doesn't this only show the latest version? `npm ls` is the one I've used to show the package registry version.
Or do I have that backwards?

scripts/npm-safe-publish.sh
```diff
@@ -104,23 +104,34 @@ fi
...
-  # Deploy logic
-  echo "Publishing $DEPRECATED_PACKAGE@$NPM_LS to NPM."
-  npm publish --otp=$OTP
+    # Backup the original package.json
+    cp package.json package.json.bak
 
-  # Restore the original package.json (trap will take care of this if the script exits prematurely)
-  mv package.json.bak package.json
+    # Ensure the original package.json is restored even if the script exits prematurely
+    trap 'mv package.json.bak package.json' EXIT
```
JamieSinn, line 119: Does this work on Linux? AFAIK this might be an OSX specific shell macro

jonathannorris: ChatGPT says so:

```Yes, the trap command works in Ubuntu Linux and many other Linux distributions. trap is a function built into the Bash shell (which Ubuntu uses) that allows you to specify commands that will be executed when the shell receives specific signals.```

**Review by JamieSinn — APPROVED** [review]

File: lib/shared/types/project.json [hunk 5]
```diff
@@ -24,6 +24,7 @@
             "outputs": ["{options.outputPath}"],
             "options": {
                 "format": ["cjs"],
+                "external": "all",
```
jonathannorris, line 27: @ajwootto fix for types lib bundling external dependencies.

PR #537, opened 2023-07-31
feat: add flushEventQueueSize and maxEventQueueSize options to control JS SDK event queue behaviour

jonathannorris:
Adding `flushEventQueueSize` and `maxEventQueueSize` DevCycleOptions to limit the size the event queue can grow to for the JS Client SDKs while also forcing a flush to happen when the event queue grows to the `flushEventQueueSize`  size. Also, the event queue is batched up into chunked requests of `100` events, the same chunk size used on the Node SDK.


jonathannorris:
* main
  * **PR #537** <a href="https://app.graphite.dev/github/pr/DevCycleHQ/js-sdks/537" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/DevCycleHQ/js-sdks/537?utm_source=stack-comment).

sdk/js/src/types.ts
```diff
@@ -93,6 +93,16 @@ export interface DevCycleOptions {
...
      */
     disableCustomEventLogging?: boolean
+
+    /**
+     * Controls the maximum size the event queue can grow to until a flush is forced. Defaults to `100`.
+     */
+    flushEventQueueSize?: number
+
+    /**
+     * Controls the maximum size the event queue can grow to until events are dropped. Defaults to `2000`.
```
suthar26, line 103: in the file, it defaults to 1000, should change to 1000
```suggestion
     * Controls the maximum size the event queue can grow to until events are dropped. Defaults to `1000`.
```

sdk/js/src/EventQueue.ts
```diff
@@ -74,31 +101,40 @@ export class EventQueue {
...
+                if (res.status !== 201) {
+                    this.eventQueue.push(...eventRequest)
+                } else {
+                    this.client.logger.info(
+                        `DVC Flushed ${eventRequest.length} Events.`,
+                    )
+                }
+            } catch (ex) {
+                this.client.eventEmitter.emitError(ex)
+                this.eventQueue.push(...eventRequest)
```
chris-hoefgen, line 123: so if the code fails to publish the events, the data goes back into the queue for a retry on a future flushEvents() call and we continue processing the other chunks. Any issues with events getting out of order or are they all fully timestamped so that they will resync in snowflake?

Also is there an potential for bad data to continually cause those events to fail and just keep getting requeued forever?

chris-hoefgen (approved):
Looks good, generally matches the pattern I've seen in the other SDKs. threw in a question but nothing blocking the PR

sdk/js/src/EventQueue.ts
jonathannorris: yea they are time-stamped when they are created. any yea failed events are a possible problem. let me investigate that.

chris-hoefgen: to be fair, this might be an issue with all of our SDKs that use a similar pattern. i think we generally report failed payloads back to WASM and it just re-queues them

jonathannorris: in the Node SDK we have a whole complicated payload retry system. However for the JS SDK I think I'm just going to say we drop `4XX` errors and retry `5XX` errors.

PR #540, opened 2023-08-03
feat: add react-native-expo package

nikolasleblanc:
`nx run react-native-expo:build:expo`

Adds a new package, `react-native-expo`, that is mostly a duplicate of `react-native`, but it triggers new `build:es5` targets in each of the react-native sdk's dependencies, via an executor that then bundles the es5 transpiled dependencies into the package.

Ideally this doesn't remain necessary for long, objective here is to not compromise the react-native sdk, and to allow us to make expo related adjustments as they become necessary.

- DVC-5386

### DIFF
--- a/sdk/react-native-expo/src/ReactNativeCacheStore.ts
+++ b/sdk/react-native-expo/src/ReactNativeCacheStore.ts
@@ -1,0 +1,26 @@
+import AsyncStorage from '@react-native-async-storage/async-storage'
+import { DVCStorage } from '@devcycle/js-client-sdk'
+
+export class ReactNativeStore implements DVCStorage {
+    store: typeof AsyncStorage
+
+    constructor() {
+        this.store = AsyncStorage
+    }
+
+    save(storeKey: string, data: unknown): void {
+        this.store.setItem(storeKey, JSON.stringify(data))
+    }
+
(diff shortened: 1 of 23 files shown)


sdk/react-native-expo/src/ReactNativeCacheStore.ts
```diff
@@ -0,0 +1,26 @@
+import AsyncStorage from '@react-native-async-storage/async-storage'
+import { DVCStorage } from '@devcycle/js-client-sdk'
+
+export class ReactNativeStore implements DVCStorage {
```
ajwootto, line 4: if these are copy pasted from the other SDK maybe just symlink them?

suthar26 (approved):

sdk/react-native-expo/src/ReactNativeCacheStore.ts
chantalchung: image
when I symlink the files I run into this error. do you know how I can update the tsconfig to allow this? I've tried asking chatGPT and this was suggested but didn't fix it

> Review the tsconfig.json file in the react-native-expo package and ensure that it is correctly configured. Make sure that the rootDir and baseUrl settings are set to the appropriate paths. The baseUrl should point to the root of the react-native-expo package, and rootDir should encompass all source files.

PR #544, opened 2023-08-14
feat: add overrides support to bucketing and add tests

ajwootto:
- support overrides in both JS and WASM libs for bucketing
- add additional optional overrides param to JS
- add new method with overrides param to WASM lib
- tests


lib/shared/bucketing-assembly-script/assembly/managers/eventQueueManager.ts
```diff
@@ -121,6 +121,7 @@ export function queueEvent(
         _getConfigData(sdkKey),
         dvcUser,
         _getClientCustomData(sdkKey),
+        null,
```
kaushalkapasi, line 124: if this is always `null` does that mean the `feature: variation` ids will be skipped from the featureVars map for overrides / self-targeting features?

ajwootto: technically yeah, but the event queue logic is only used by serverside SDKs, which currently would not deal with overrides directly. The only thing that matters is the code used by the bucketing API, which is just the config generator  portion.

kaushalkapasi (approved):

lib/shared/bucketing-assembly-script/assembly/bucketing/bucketing.ts
```diff
@@ -196,38 +229,77 @@ export function bucketUserForVariation(
...
+
+            if (variation == null) {
+                continue
+            }
+
+            bucketedFeatureVariations.push({ feature, variation })
+        }
+    }
+
+    for (let i = 0; i < bucketedFeatureVariations.length; i++) {
```
jonathannorris, line 286: Ideally, we could structure this to not have to write to `bucketedFeatureVariations` and iterate over it again, as we are adding some allocations and needing to loop over this new array.

ajwootto: updated

lib/shared/bucketing-assembly-script/assembly/types/configBody.ts
```diff
@@ -89,64 +88,107 @@ export class ConfigBody {
...
+    static fromString(
+        configStr: string,
+        etag: string | null = null,
+    ): ConfigBody {
         const configJSON = JSON.parse(configStr)
-        if (!configJSON.isObj) throw new Error('generateBucketedConfig config param not a JSON Object')
+        if (!configJSON.isObj)
+            throw new Error(
+                'generateBucketedConfig config param not a JSON Object',
+            )
```
jonathannorris, line 114:
```suggestion
        if (!configJSON.isObj) {
            throw new Error(
                'generateBucketedConfig config param not a JSON Object',
            )
          }
```

jonathannorris, line 114:
```suggestion
        if (!configJSON.isObj) {
            throw new Error(
                'generateBucketedConfig config param not a JSON Object',
            )
        }
```

lib/shared/bucketing-assembly-script/assembly/types/configBody.ts
```diff
@@ -184,4 +226,10 @@ export class ConfigBody {
             ? this._variableIdToFeatureMap.get(variable_id)
             : null
     }
+
+    getFeatureForId(feature_id: string): Feature | null {
```
jonathannorris, line 230: What is this used for?

jonathannorris (changes requested):

jonathannorris (approved):

PR #556, opened 2023-09-11
ci: Add a skeleton release workflow to allow testing on GA

rob-odwyer:
This just allows testing the workflow on a branch later - the file needs to be merged first.


chris-hoefgen (approved):
Not much to review, i assume you just want this approved so you can keep moving forward with the meaty bits?

PR #557, opened 2023-09-12
fix: Defer React SDK initialization on server

emmawillis:
- Always default isInitialized state to false in DevCycleProvider, so that it won't become true until after the client is created 
- Defer client initialization on server side so dvc client will only be initialized on the client side (server rendering will always be loading state, which will match initial client render and prevent hydration errors) 
- Use a ref to wrap the current dvc client on each DevCycleProvider, so that there is a separate client created for each instance of DevCycleProvider (rather than one client shared across the server)

### DIFF
--- a/sdk/react/src/DevCycleProvider.tsx
+++ b/sdk/react/src/DevCycleProvider.tsx
@@ -9,16 +9,12 @@ type Props = {
     children: ReactNode
 }
 
-let client: DevCycleClient | undefined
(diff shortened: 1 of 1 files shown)


ajwootto (commented):
looks fine but i'll let someone else review it

jsalaber (approved):

kaushalkapasi (approved):

PR #559, opened 2023-09-15
ci: Implement a basic release workflow (DVC-8825)

rob-odwyer:
This ports the manual deployment flow described [in the repo here](https://github.com/DevCycleHQ/js-sdks/blob/main/README.md#publishing-a-release) to a Github Actions workflow. Because we can't prompt the user for individual version increment types for each package, it's limited to one type of version increment (major/minor/patch) for all affected packages.

It also adds the ability to force a version bump to every package.


.github/workflows/release.yml
```diff
@@ -29,10 +29,70 @@ permissions:
...
+              run: |
+                  git config --global user.email "github-tracker-bot@taplytics.com"
+                  git config --global user.name "DevCycle Automation"
+
+            - name: Use Node.js ${{ matrix.node-version }}
+              uses: actions/setup-node@v3
+              with:
+                  node-version: ${{ matrix.node-version }}
+                  cache: 'yarn'
+
```
JamieSinn, line 60: You'll have to add the registry url here:
https://docs.github.com/en/actions/publishing-packages/publishing-nodejs-packages#publishing-packages-to-the-npm-registry

scripts/lerna-version-ci.sh
```diff
@@ -0,0 +1,113 @@
...
+if [ -z "$AFFECTED_PROJECTS" ]; then
+  AFFECTED_PROJECTS=$(yarn nx print-affected --base $LAST_TAGGED_SHA --select=projects)
+fi
+
+echo "Affected projects: $AFFECTED_PROJECTS"
+
+# exit if no affected projects
+if [ -z "$AFFECTED_PROJECTS" ]; then
+  echo "No affected projects found. Exiting."
+  exit 1
```
JamieSinn, line 56: If there are no affected projects - do we want this to crash the github action?

scripts/npm-safe-publish.sh
```diff
@@ -67,35 +92,30 @@ if [[ "$NPM_SHOW" != "$NPM_LS" ]]; then
...
-    npm publish --local
+    echo "[DRY RUN] Not publishing $PACKAGE@$NPM_LS to NPM."
   fi
+
+  if [[ "$?" != 0 ]]; then
+    echo "Publish failed. Aborting."
+    exit 1
+  fi
+
+  curl -X POST \
```
JamieSinn, line 115: We can remove all the sleuth stuff since this is now able to be tracked as a github action in Sleuth directly.

.github/workflows/release.yml
```diff
@@ -29,10 +29,70 @@ permissions:
...
+            - name: Build UMD bundles
+              run: |
+                  yarn nx build:cdn js
+              env:
+                  NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+            - name: Upload CDN Release
+              run: aws s3 sync dist/sdk/js s3://js.devcycle.com/ --include "devcycle.min.js*" --acl public-read
+              if: inputs.draft != true && github.ref == 'refs/heads/main'
```
JamieSinn, line 98: We could probably combine these into one step too if needed

.github/workflows/release.yml
rob-odwyer: ah, I wonder if that's why the npm auth was working for me locally but not on GHA.

scripts/lerna-version-ci.sh
rob-odwyer: I think so, because if it silently does nothing you'll think you released something when you didn't. There's no downside I can see to reporting an error here.

JamieSinn: As long as we're consistent on the no-op = abort for other releases across the other sdks

.github/workflows/release.yml
rob-odwyer: It's been helpful to have them split for testing that the CDN build works without actually pushing anything.

scripts/lerna-version-ci.sh
rob-odwyer: I don't think the same thing applies for the other non-monorepo SDKs. Because there's no need to select a subset of projects to push updates for, the idea of pushing a no-op release is reasonable. With this flow, if you want to push a dummy update you can select the "Force update all packages" option.

JamieSinn: I was more thinking about the C# SDK - which currently allows you to versionbump+release a no-op

JamieSinn: But with the force update - I think that covers it !

scripts/npm-safe-publish.sh
rob-odwyer: Leaving this in for now because I don't understand the sleuth configuration, and it's apparently working the way it is.

JamieSinn: Frankly neither do I, I just get Nik to do it

JamieSinn (approved):

PR #563, opened 2023-09-25
fix: instantiate wasm every time instead of using a global singleton (DVC-9129)

rob-odwyer:
This allows the SDK to be instantiated multiple times for the same SDK key, e.g. for hot reloading.

### DIFF
--- a/sdk/nodejs/src/bucketing.ts
+++ b/sdk/nodejs/src/bucketing.ts
@@ -15,7 +15,7 @@ export const importBucketingLib = async ({
         bucketingInstantiationCount++
         if (bucketingInstantiationCount > bucketingInstantiationWarning) {
             logger?.warn(
-                `Warning: The DevCycle SDK has been instantiated over ${bucketingInstantiationWarning} times. ` +
+                `Warning: The DevCycle SDK has been instantiated ${bucketingInstantiationWarning} times. ` +
                     'This may cause higher than expected memory usage and indicate a faulty integration.',
             )
         }
--- a/sdk/nodejs/src/bucketing.ts
+++ b/sdk/nodejs/src/bucketing.ts
@@ -3,22 +3,25 @@ import { DVCLogger, DVCReporter } from '@devcycle/types'
 import { DevCycleOptions } from './types'
 
 let Bucketing: Exports | null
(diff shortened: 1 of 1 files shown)


sdk/nodejs/src/bucketing.ts
```diff
@@ -3,22 +3,25 @@ import { DVCLogger, DVCReporter } from '@devcycle/types'
...
-        await InstantiatePromise
-        return
-    }
     const debugWASM = process.env.DEVCYCLE_DEBUG_WASM === '1'
-    InstantiatePromise = instantiate(debugWASM).then((exports) => {
+    await instantiate(debugWASM).then((exports) => {
+        bucketingInstantiationCount++
+        if (bucketingInstantiationCount > bucketingInstantiationWarning) {
+            logger?.warn(
+                `Warning: The DevCycle SDK has been instantiated over ${bucketingInstantiationWarning} times. ` +
```
jonathannorris, line 18:
```suggestion
                `Warning: The DevCycle SDK has been instantiated ${bucketingInstantiationWarning} times. ` +
```

sdk/nodejs/src/bucketing.ts
```diff
@@ -3,22 +3,25 @@ import { DVCLogger, DVCReporter } from '@devcycle/types'
...
     logger,
     options,
 }: { logger?: DVCLogger; options?: DevCycleOptions } = {}): Promise<void> => {
-    if (InstantiatePromise) {
-        await InstantiatePromise
-        return
-    }
     const debugWASM = process.env.DEVCYCLE_DEBUG_WASM === '1'
-    InstantiatePromise = instantiate(debugWASM).then((exports) => {
+    await instantiate(debugWASM).then((exports) => {
```
jonathannorris, line 14: Can we add tests to make sure things work with multiple WASM libs? Probably also makes sense to add a test harness tests to make sure multiple clients work.

jonathannorris: yea for sure I much prefer reviewing a draft PR, really helps clarify what this change would mean. But yea in general I think this is a good idea.

rob-odwyer:
Closing this for the time being - it's more complicated than I hoped to create a new WASM instance every time and will require some refactoring.

PR #561, opened 2023-09-20
feat: DVC-9055 adding Cloudflare Worker example app for new js-cloud-server-sdk

jonathannorris:

PR #560, opened 2023-09-20
feat: DVC-9055 extract JS Cloud Bucketing SDK from NodeJS SDK

jonathannorris:
Extracting the JS Cloud Bucketing SDK from the NodeJS SDK, this will allow for the more generic JS Cloud Bucketing SDK to be used in places where the NodeJS SDK couldn't be (Cloudflare Workers / Edge Worker Runtimes / Electron Apps / etc). The NodeJS SDK will import this new SDK for its existing `DevCycleCloudClient` interface. To avoid duplication of code, most of the shared logic from the NodeJS SDK has been moved to the new JS Cloud Bucketing SDK and imported from there.

PR #564, opened 2023-09-26
feat: DVC-9059 pull out EnvironmentConfigManager to its own library, change nodejs build to define external pacakges

jonathannorris:
- Update NodeJS build process to bundle libraries that are not defined in the `package.json` / `external` under the `build` step.
- Pull out `EnvironmentConfigManager` into a generic lib to be used by the Edge Worker SDK
- Other libs like the server request handling will be pulled into libs with further stacked PRs.


jonathannorris:
* main
  * **PR #560** <a href="https://app.graphite.dev/github/pr/DevCycleHQ/js-sdks/560" target="_blank">Graphite</a> 
    * **PR #561** <a href="https://app.graphite.dev/github/pr/DevCycleHQ/js-sdks/561" target="_blank">Graphite</a> 
      * **PR #564** <a href="https://app.graphite.dev/github/pr/DevCycleHQ/js-sdks/564" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/DevCycleHQ/js-sdks/564?utm_source=stack-comment).

sdk/nodejs/project.json
```diff
@@ -1,70 +1,85 @@
...
+        "packageJson": "sdk/nodejs/package.json",
+        "buildableProjectDepsInPackageJsonType": "dependencies",
+        "main": "sdk/nodejs/src/index.ts",
+        "assets": [
+          "sdk/nodejs/*.md"
+        ],
+        "external": [
+            "shared-types",
+            "shared-bucketing-test-data",
+            "js-cloud-server-sdk"
```
ajwootto, line 65: assembly script?

lib/shared/config-manager/src/index.ts
```diff
@@ -17,12 +21,18 @@ export class EnvironmentConfigManager {
     private readonly requestTimeoutMS: number
     private readonly cdnURI: string
     fetchConfigPromise: Promise<void>
-    private intervalTimeout?: NodeJS.Timeout
+    private intervalTimeout?: any
```
ajwootto, line 24: why does this need to be any?

lib/shared/config-manager/src/index.ts
```diff
@@ -17,12 +21,18 @@ export class EnvironmentConfigManager {
...
     private disablePolling = false
+    private readonly setConfigBuffer: SetConfigBuffer
+    private readonly setInterval: SetIntervalInterface
+    private readonly clearInterval: ClearIntervalInterface
 
     constructor(
         logger: DVCLogger,
         sdkKey: string,
+        setConfigBuffer: SetConfigBuffer,
+        setInterval: SetIntervalInterface,
```
ajwootto, line 34: this polling config fetcher seems very specific to the nodejs local bucketing SDK, what purpose would it serve in the worker SDK other than to fetch it once?

lib/shared/config-manager/src/index.ts
jonathannorris: basically because `NodeJS.Timeout` isn't supported in webworker / browser environments, they just return a number I believe. So the strategy here was to add a `setInterval` + `clearInterval` interface that would be passed in, so this needs to be an `Any` if we want it to compile when used in the non-NodeJS SDKs.

lib/shared/config-manager/src/index.ts
jonathannorris: workers do last for a while (I think I've seen 10's of minutes if not longer), we'd still want polling for them.

lib/shared/config-manager/__tests__/environmentConfigManager.spec.ts
```diff
@@ -1,23 +1,40 @@
...
+import { DVCLogger } from '@devcycle/types'
 import { getEnvironmentConfig } from '../src/request'
 
 const setInterval_mock = mocked(setInterval)
 const getEnvironmentConfig_mock = mocked(getEnvironmentConfig)
 const logger = dvcDefaultLogger()
 
+const mockSDKConfig = jest.fn()
+
+function getConfigManager(
```
jonathannorris, line 22: Moving test to new config-manager lib

lib/shared/config-manager/src/index.ts
```diff
@@ -136,7 +150,8 @@ export class EnvironmentConfigManager {
             )
         } else if (responseError?.status === 403) {
             this.stopPolling()
-            throw new UserError(`Invalid SDK key provided: ${this.sdkKey}`)
+            // UserError
```
jonathannorris, line 153: This gets re-added in one of the next stacked PRs.

lib/shared/config-manager/src/request.ts
```diff
@@ -0,0 +1,38 @@
+import { RequestInitWithRetry } from 'fetch-retry'
+import { get } from '@devcycle/js-cloud-server-sdk'
+
+export async function getEnvironmentConfig(
```
jonathannorris, line 4: move the `request.ts` methods only used by the config-manager here.

sdk/nodejs/package.json
```diff
@@ -1,15 +1,18 @@
...
-    }
+  "name": "@devcycle/nodejs-server-sdk",
+  "version": "1.15.4",
+  "description": "The DevCycle NodeJS Server SDK used for feature management.",
+  "license": "MIT",
+  "devDependencies": {
+    "@devcycle/bucketing-test-data": "^1.2.21"
+  },
+  "dependencies": {
+    "@devcycle/bucketing-assembly-script": "^1.8.3",
```
jonathannorris, line 10: Due to the changed to use the `external` option as part of the build, we now need to manually define our dependencies for when this will be published to NPM.

sdk/nodejs/project.json
```diff
@@ -1,70 +1,85 @@
...
+      "options": {
+        "outputPath": "dist/sdk/nodejs",
+        "tsConfig": "sdk/nodejs/tsconfig.lib.json",
+        "packageJson": "sdk/nodejs/package.json",
+        "buildableProjectDepsInPackageJsonType": "dependencies",
+        "main": "sdk/nodejs/src/index.ts",
+        "assets": [
+          "sdk/nodejs/*.md"
+        ],
+        "external": [
```
jonathannorris, line 62: `external` packages now need to be defined here, all other packages will be compiled into the SDK.

sdk/nodejs/src/bucketing.ts
```diff
@@ -55,3 +55,11 @@ export const getBucketingLib = (): Exports => {
 export const cleanupBucketingLib = (): void => {
     Bucketing = null
 }
+
+export const setConfigDataUTF8 = (
```
jonathannorris, line 59: externalizing NodeJS only functionality of the `EnvironmentConfigManager` class

sdk/nodejs/src/client.ts
```diff
@@ -73,6 +77,9 @@ export class DevCycleClient {
                 this.configHelper = new EnvironmentConfigManager(
                     this.logger,
                     sdkKey,
+                    setConfigDataUTF8,
+                    setInterval,
+                    clearInterval,
```
jonathannorris, line 82: The interface for these is different between NodeJS / Web, we need to pass it in as an interface.

lib/shared/config-manager/src/index.ts
ajwootto: this would affect the behaviour of the SDK, not sure how you can safely make this change even if its put back later

lib/shared/config-manager/src/index.ts
ajwootto: if you wanted to get fancy you could use a type generic for this. Type it so that the generic is what is returned by the SetIntervalInterface and then use the same generic here.

sdk/nodejs/project.json
```diff
@@ -1,70 +1,87 @@
...
+      "options": {
+        "outputPath": "dist/sdk/nodejs",
+        "tsConfig": "sdk/nodejs/tsconfig.lib.json",
+        "packageJson": "sdk/nodejs/package.json",
+        "buildableProjectDepsInPackageJsonType": "dependencies",
+        "main": "sdk/nodejs/src/index.ts",
+        "assets": [
+          "sdk/nodejs/*.md"
+        ],
+        // External pacakges now need to be defined here, all other pacakges will be compiled into the sdk.
```
ajwootto, line 62: technically you cant add comments to json files

jonathannorris: yea I know, but also it works with them... so I thought it would be helpful to try and explain what external packages are (one of the reasons JSON isn't a good format for config files).

jsalaber (approved):

lib/shared/config-manager/src/index.ts
jonathannorris: From what I can see it doesn't actually affect the behaviour of the SDK, and will be re-added in the next PR where it it is moved to the new request lib.

jonathannorris: I can manually copy it in here temporarily.

lib/shared/config-manager/project.json
```diff
@@ -0,0 +1,41 @@
+{
+    "name": "config-manager",
+    "$schema": "../../../node_modules/nx/schemas/project-schema.json",
+    "sourceRoot": "lib/shared/config-manager/src",
+    "projectType": "library",
+    "targets": {
+        "build": {
```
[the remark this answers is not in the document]
jonathannorris: good point, let me see if the tests or anything break if I remove it.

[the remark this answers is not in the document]
jonathannorris: looks like that worked.

ajwootto (approved):

PR #569, opened 2023-10-17
feat: add option for bootstrapping config in JS SDK

ajwootto:
- add bootstrapConfig option to provide initial config to JS SDK

### DIFF
--- a/sdk/js/__tests__/Client.spec.js
+++ b/sdk/js/__tests__/Client.spec.js
@@ -56,6 +56,40 @@ describe('DevCycleClient tests', () => {
         expect(client.config).toStrictEqual(testConfig)
     })
 
+    it('should use bootstrapped config when set, not fetch a new one, initialize SDK', async () => {
+        const client = new DevCycleClient(
+            'test_sdk_key',
+            { user_id: 'user1' },
+            { bootstrapConfig: testConfig },
+        )
+        await client.onInitialized
+        expect(getConfigJson_mock).not.toBeCalled()
+        expect(client.config).toStrictEqual(testConfig)
+        expect(client.user.user_id).toEqual('user1')
+    })
+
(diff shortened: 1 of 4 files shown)


sdk/js/__tests__/Client.spec.js
```diff
@@ -56,6 +56,40 @@ describe('DevCycleClient tests', () => {
...
+        // client
+        const client2 = new DevCycleClient(
+            'test_sdk_key',
+            { user_id: 'user1' },
+            { bootstrapConfig: testConfig },
+        )
+        await client2.onInitialized
+        // one more call to get anon ID
+        expect(spy).toHaveBeenCalledTimes(3)
+        expect(client2.config).toStrictEqual(testConfig)
```
suthar26, line 90: if the getConfigJson_mock is using a little changed config, we can test to make sure both clients use different configs

ajwootto: that's basically what I'm testing here, that client2 has the config provided for boostrapping rather than the modified one that client1 got

suthar26: i missed the `cached: true` 😅

suthar26 (approved):

sdk/js/src/Client.ts
```diff
@@ -43,6 +43,7 @@ type variableEvaluatedHandler = (
 
 export type DevCycleOptionsWithDeferredInitialization = DevCycleOptions & {
     deferInitialization: true
+    bootstrapConfig?: never
```
jonathannorris, line 46: I've never seen `never` before...

jonathannorris: https://www.tutorialsteacher.com/typescript/typescript-never

jonathannorris (approved):

PR #590, opened 2023-10-27
fix: Use populated user when checking if user is anonymous

laurawarr:
In [this PR](https://github.com/DevCycleHQ/js-sdks/commit/beed17838888d8aedc53e520721fa5d3a934ec56#diff-7e88ae7eb811449dfe4a792952dab7d8accab565833b6a8a5764488a232d4325R191) there was a change that switched from using the populated user to using the initial user to check if the user is anonymous. 

`isAnonymous` would only be true on the initial user if the SDK is explicitly called with this property. The populated user will have isAnonymous set to true if a user id is not specified.

This would cause the cached anonymous user ID to be cleared on every initialization, resulting in more MAUs than expected.

### DIFF
--- a/sdk/js/src/Client.ts
+++ b/sdk/js/src/Client.ts
@@ -216,7 +216,7 @@ export class DevCycleClient<
 
         this.eventEmitter.emitInitialized(true)
 
-        if (initialUser.isAnonymous) {
+        if (this.user.isAnonymous) {
             this.store.saveAnonUserId(this.user.user_id)
         } else {
             this.store.removeAnonUserId()
--- a/sdk/js/__tests__/Client.spec.js
+++ b/sdk/js/__tests__/Client.spec.js
@@ -233,6 +233,19 @@ describe('DevCycleClient tests', () => {
         expect(window.localStorage.getItem(StoreKey.AnonUserId)).toBeNull()
     })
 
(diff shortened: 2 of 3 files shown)


sdk/js/src/Client.ts
```diff
@@ -216,7 +216,7 @@ export class DevCycleClient<
 
         this.eventEmitter.emitInitialized(true)
 
-        if (initialUser.isAnonymous) {
```
ajwootto, line 219: i thought it was the intention that the user passed into the SDK needed to either have isAnonymous=true, or have a user id set, otherwise you'd receive an error? That's what the comments on the DevCycleUser interface say anyway

laurawarr: It looks like that's an old comment that wasn't removed when this change was made. After anonymous user ID caching was implemented, if the user ID isn't passed then it's assumed the user is anonymous https://github.com/DevCycleHQ/js-sdks/blob/3ee5528956a09afae75c921fcbc9521b15fff551/sdk/js/src/User.ts#L52

UofPx has been using the SDK with an empty user object, and as a result of this each initialization clears the ID cache and counts as an MAU

ajwootto: ok. Can you fix that comment and maybe add a test for this?

ajwootto (approved):

PR #600, opened 2023-11-06
fix: Remove protoc installation in workflows

rob-odwyer:
This step takes a few minutes on every run, but I'm pretty sure it's no longer required.

### DIFF
--- a/.github/workflows/nx-affected-test.yml
+++ b/.github/workflows/nx-affected-test.yml
@@ -17,8 +17,6 @@ jobs:
               with:
                   node-version: ${{ matrix.node-version }}
                   cache: 'yarn'
-            - name: Install protoc for protobuf compilation
-              run: sudo apt-get install -y protobuf-compiler
             - run: git fetch origin main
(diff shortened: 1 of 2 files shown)


JamieSinn:
I think this is still used as part of building the protobuf schema here:
https://github.com/DevCycleHQ/js-sdks/blob/main/lib/shared/bucketing-assembly-script/project.json#L106

rob-odwyer:
@JamieSinn Right, but that's not called by the main build anywhere, and the resulting compiled files are checked into the repo.
- [Manual step for rebuilding them is described here](https://github.com/DevCycleHQ/js-sdks/blob/main/lib/shared/bucketing-assembly-script/README.md#compile-protobuf-files)
- [Compiled files are checked in here](https://github.com/DevCycleHQ/js-sdks/tree/main/lib/shared/bucketing-assembly-script/protobuf)

JamieSinn:
Ah right okay; wasn't sure if that was now a manual process or not.

PR #658, opened 2023-12-21
fix: deferred mode check and add tests

ajwootto:
- fix logic checking for deferred mode options to explicitly look for true
- add tests around this

### DIFF
--- a/sdk/js/src/index.ts
+++ b/sdk/js/src/index.ts
@@ -38,6 +38,41 @@ export type DVCOptionsWithDeferredInitialization =
 export type { DevCycleOptionsWithDeferredInitialization }
 export { DevCycleClient }
 
+const determineUserAndOptions = (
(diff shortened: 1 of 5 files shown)


sdk/js/src/index.ts
```diff
@@ -38,6 +38,41 @@ export type DVCOptionsWithDeferredInitialization =
...
+          isDeferred: true
+      }
+    | {
+          user: DevCycleUser
+          options: DevCycleOptions
+          isDeferred: false
+      } => {
+    let user: DevCycleUser | undefined = undefined
+    if (!!userOrOptions && 'deferInitialization' in userOrOptions) {
+        if (userOrOptions.deferInitialization) {
```
jsalaber, line 57: is this if statement redundant?

ajwootto: no, because it's trying to catch the case where deferInitialization is explicitly set but it is set to false

jsalaber (approved):

PR #664, opened 2024-01-03
feat: add missing methods for SDK

ajwootto:
- add useTrack, useAllVariables, useAllFeatures
- update readme with clearer setup instructions

PR #665, opened 2024-01-04
fix: streaming in cases where the promise resolves before the layout is rendered

ajwootto:
- sometimes the initialize promise will resolve before the layout or other parts of the page are rendered
- in this case the previous implementation of streaming would cause hydration errors because the server would render its contents with the variable values, while the client would wait for the promise and re-render on the next pass after realizing it was resolved
- to fix this, the server basically needs to wait for the promise in a non-blocking way and then notify the client during the first render whether it is safe to `use` the promise because it has already resolved, thus receiving its resolved contents on the first render pass without blocking anything in streaming mode
- this also removes the requirement to await the getClientContext method, and brings the waiting of the user's identity from the provided getter inside the streaming system. That means that the async process to determine the user's identity does not need to block rendering either, it can be part of the streaming process

### DIFF
--- a/e2e/nextjs/app-router/app/app/layout.tsx
+++ b/e2e/nextjs/app-router/app/app/layout.tsx
@@ -16,7 +16,7 @@ export default async function RootLayout({
     return (
         <html lang="en">
             <body>
-                <DevCycleClientsideProvider context={await getClientContext()}>
+                <DevCycleClientsideProvider context={getClientContext()}>
                     {children}
                 </DevCycleClientsideProvider>
             </body>
--- a/e2e/nextjs/app-router/app/app/page.tsx
+++ b/e2e/nextjs/app-router/app/app/page.tsx
@@ -3,7 +3,7 @@ import { ServerComponent } from '@/app/ServerComponent'
 import React, { Suspense } from 'react'
 import Link from 'next/link'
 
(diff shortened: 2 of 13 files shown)


ajwootto:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/DevCycleHQ/js-sdks/665?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* `main`
  * **PR #664** <a href="https://app.graphite.dev/github/pr/DevCycleHQ/js-sdks/664?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #665** <a href="https://app.graphite.dev/github/pr/DevCycleHQ/js-sdks/665?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/DevCycleHQ/js-sdks/665?utm_source=stack-comment).

jonathannorris (approved):

PR #678, opened 2024-01-09
fix: NextJS example apps to use new SDK interface

jonathannorris:


jonathannorris:
* `main`
  * **PR #678** <a href="https://app.graphite.dev/github/pr/DevCycleHQ/js-sdks/678?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/DevCycleHQ/js-sdks/678?utm_source=stack-comment).

examples/nextjs/app-router/app/ServerIdentity.tsx
```diff
@@ -17,7 +17,7 @@ export const ServerIdentity = async function () {
             </span>
             <b>Server Identity</b>
             <span data-testid={'server-user-id'}>
-                {getUserIdentity()?.user_id}
+                {(await getUserIdentity()).user_id}
```
ajwootto, line 20: this isn't really the intention of this line. It was meant to reflect what the user identity the SDK currently knows it has. It's calling a method that's still exported from the SDK, but actually wouldn't work if it was called so maybe we should remove it or make it work

examples/nextjs/app-router/app/layout.tsx
```diff
@@ -10,21 +11,9 @@ export default async function RootLayout({
...
-                    user={{
-                        user_id:
-                            process.env.NEXT_PUBLIC_USER_ID || 'server-user',
-                    }}
-                    options={{
-                        enableStreaming:
-                            process.env.NEXT_PUBLIC_ENABLE_STREAMING === '1',
-                    }}
-                >
+                <DevCycleClientsideProvider context={await getClientContext()}>
```
ajwootto, line 14: this isnt async anymore

examples/nextjs/app-router/app/ServerIdentity.tsx
jonathannorris: I could set the `user_id` as a global in the `devcycle.tsx` file, but I couldn't see a way to get the user_id from the context that is returned.

ajwootto: yeah there currently isn't a way, we'd have to return the `getUserIdentity` function from here

jonathannorris: why I can't I just return it from the `devcycle.tsx` file like it is right now?

ajwootto: you can but that wasn't the point of what it was doing before. It was printing what the SDK knows the user's identity to be

ajwootto: well originally it was saying "here is what the user's identity is currently set to within the SDK" and now its saying "here is what I'm telling the SDK the user should be set to". The result is the same but the place where you're getting the info from is different. However I think we should just remove the ability to `getUserIdentity` from the SDK for now, so this is probably fine

ajwootto (approved):

PR #692, opened 2024-01-16
fix: eliminate a bunch of the shared context stuff and move the user agent call

ajwootto:
- refactor to remove a bunch of the React cache calls that were sharing data easily passed around
- make sure each call to `getVariableValue` gets its own Promise rather than using a shared one. That promise will resolve immediately if things are cached already
- add a delay to the streaming e2e app user getter
- add a test looking for streaming loading state


sdk/nextjs/src/server/initialize.ts
```diff
@@ -14,28 +14,47 @@ const jsClientOptions = {
...
     options: DevCycleNextOptions = {},
-): Promise<Awaited<ReturnType<typeof getDevCycleServerData>>> => {
-    setSDKKey(sdkKey)
-    setOptions(options)
-    const user = await userGetter()
+): Promise<DevCycleServerData> => {
+    validateSDKKey(sdkKey)
+    // TODO moving this call to inside `getBucketedConfig` appears to cause static build issues from reading headers
+    // Might be a bug in Next, if moving this make sure to verify you can `yarn next build` the e2e app router app
+    const userAgent = getUserAgent(options)
```
ajwootto, line 33: moving this here seems to fix the static build issue..

laurawarr (approved):

PR #704, opened 2024-01-24
feat: add IfEnabled helper

ajwootto:
Add a helper for dynamically importing components into the client bundle depending on feature flag values.

It's used like this:

```typescript
'use client'
import {
    ifEnabled,
} from '@devcycle/nextjs-sdk'

const ConditionalComponent = ifEnabled(
    'enabled-feature',
    () => import('./ConditionalClientComponent'),
)

export const ClientComponent = () => {
    return (
        <div>
            <ConditionalComponent />
        </div>
    )
}
```
The result is that if the flag evaluates to false, the component code is never sent to the client.

If it evaluates to true, the initial rendered content is send by the server and the component is "progressively enhanced" with the interactive JS after the page loads (lazy loaded)


sdk/nextjs/src/client/IfEnabled.tsx
```diff
@@ -0,0 +1,18 @@
+import { ComponentProps, ComponentType } from 'react'
+import dynamic from 'next/dynamic'
+import useVariableValue from './useVariableValue'
+
+// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
+export const ifEnabled = <T extends { default: ComponentType<any> }>(
```
jsalaber, line 6: is there a better name than `ifEnabled`, maybe `renderIfEnabled`?

sdk/nextjs/src/server/setupDevCycle.tsx
```diff
@@ -6,6 +6,7 @@ import { getUserAgent } from './userAgent'
 import { getAllVariables } from './getAllVariables'
 import { getAllFeatures } from './allFeatures'
 import { DevCycleNextOptions } from '../common/types'
+import { ComponentProps, ComponentType, lazy } from 'react'
```
jonathannorris, line 9: are these used? can they be cleaned up?

jsalaber (approved):

jonathannorris (approved):

PR #727, opened 2024-02-01
feat: add conditional render helper and border highlighting

ajwootto:
- add utilities for conditionally rendering a component, and swapping component implementations
- add an option to highlight the conditionally rendered component with a border that contains the variable key and value as a label, with a link to the dashboard where that variable is defined.

image

### DIFF
--- a/sdk/react/src/RenderIfEnabled.tsx
+++ b/sdk/react/src/RenderIfEnabled.tsx
@@ -1,0 +1,64 @@
+import useVariableValue from './useVariableValue'
+import { DVCVariableValue } from '@devcycle/js-client-sdk'
+
+type RenderIfEnabledProps<T extends DVCVariableValue> =
+    | {
+          children: React.ReactNode
+          variableKey: string
+          targetValue: T
+          defaultValue: T
+          showBorder?: boolean
+      }
+    | {
+          children: React.ReactNode
+          variableKey: string
+          showBorder?: boolean
+      }
+
+export const RenderIfEnabled = <T extends DVCVariableValue>(
+    props: RenderIfEnabledProps<T>,
+): React.ReactNode => {
+    let targetValue: DVCVariableValue
+    let defaultValue: DVCVariableValue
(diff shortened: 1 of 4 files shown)


jonathannorris (approved):

sdk/react/src/SwapComponents.tsx
```diff
@@ -0,0 +1,22 @@
+import { ComponentProps, ComponentType } from 'react'
+import useVariableValue from './useVariableValue'
+
+// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
+export const SwapComponents = <T extends ComponentType<any>>(
```
jonathannorris, line 5: Swap components makes me almost feel like this could swap between `n` number of different components. Maybe we could use the name of the component to match against a string value?

ajwootto: there's definitely a more complicated version  we could have where we map the possible values of variable to the component that should be rendered.

ajwootto:
* **#727** <a href="https://app.graphite.dev/github/pr/DevCycleHQ/js-sdks/727?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

 Join @ajwootto and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

sdk/react/src/RenderIfEnabled.tsx
```diff
@@ -0,0 +1,68 @@
...
+}
+
+type RenderIfEnabledProps<T extends DVCVariableValue> =
+    | CommonProps
+    | (CommonProps & {
+          targetValue: T
+          defaultValue: T
+      })
+
+export const RenderIfEnabled = <T extends DVCVariableValue>(
```
jonathannorris, line 18: Are there any unit / E2E tests we can add for these new interfaces?

jonathannorris (approved):

PR #740, opened 2024-02-21
feat: Add support for service workers by using indexedDB instead of localStorage

elliotCamblor:
- update several methods to return promises to support await-ing indexedDB connection
-  Add alternative storage strategy for saving to indexedDB rather than window.localStorage. This enables service worker implementations which don't have access to the `window` object. This strategy is invoked when the `self` global is accessible and has the name `ServiceWorkerGlobalScope` 
- consolidate DefaultStorage and DefaultCacheStorage because only one of them was being used anyway
- No longer logging a warning on initialization about not having access to the window object if its been detected that we're running in a service worker environment

How was this tested:
- I ran this against the example app (non-service worker) with both storage strategies succesfully.
- I also installed this local version of the sdk in a test chrome extension successfully (confirmed that indexedDB is being used and anon ID's are being recycled on subsequent initializations. Also configs are being cached)


lib/shared/server-request/package.json
```diff
@@ -1,5 +0,0 @@
-{
-    "name": "@devcycle/server-request",
```
jonathannorris, line 2: why is this removed?

sdk/js/src/Client.ts
```diff
@@ -469,20 +469,20 @@ export class DevCycleClient<
 
             this.onInitialized
```
jonathannorris, line 470: it's very weird we use `.then` / `.catch` here instead of `async` / `await`.

sdk/js/src/DefaultCacheStore.ts
```diff
@@ -1,26 +1,39 @@
...
-        this.store?.setItem(storeKey, JSON.stringify(data))
+    constructor(isTesting = false) {
+        super()
+        this.isTesting = isTesting
+        this.init()
+    }
+    async init(): Promise<void> {
+        this.store = this.isTesting ? stubbedLocalStorage : window.localStorage
+    }
+    async save(storeKey: string, data: unknown): Promise<void> {
```
jonathannorris, line 26:
```suggestion
    }
    
    async init(): Promise<void> {
        this.store = this.isTesting ? stubbedLocalStorage : window.localStorage
    }
    
    async save(storeKey: string, data: unknown): Promise<void> {
```

sdk/js/src/DefaultCacheStore.ts
```diff
@@ -1,26 +1,39 @@
...
+    private isTesting: boolean
 
-    save(storeKey: string, data: unknown): void {
-        this.store?.setItem(storeKey, JSON.stringify(data))
+    constructor(isTesting = false) {
+        super()
+        this.isTesting = isTesting
+        this.init()
+    }
+    async init(): Promise<void> {
```
jonathannorris, line 23: is `init()` ever called out-side of the constructor? could this just be in the constructor?

sdk/js/src/DefaultCacheStore.ts
```diff
@@ -32,5 +45,91 @@ const stubbedLocalStorage = {
     key: () => null,
     length: 0,
 }
+// IndexedDB implementation
+export class IndexedDBStrategy extends StorageStrategy {
+    override store: IDBDatabase
+    private ready: Promise<void>
+    private static storeName = 'DevcycleStore'
+    private static DBName = 'DevcycleDB'
+    constructor() {
```
jonathannorris, line 54:
```suggestion
    private static DBName = 'DevcycleDB'
    
    constructor() {
```

jonathannorris, line 54:
```suggestion
    private static storeName = 'DevCycleStore'
    private static DBName = 'DevCycleDB'
    
    constructor() {
```

sdk/js/src/DefaultCacheStore.ts
```diff
@@ -32,5 +45,91 @@ const stubbedLocalStorage = {
...
+
+    // IndexedDB remove
+    async remove(storeKey: string): Promise<void> {
+        await this.ready
+        const tx = this.store.transaction(
+            IndexedDBStrategy.storeName,
+            'readwrite',
+        )
+        const store = tx.objectStore(IndexedDBStrategy.storeName)
+        store.delete(storeKey)
```
jonathannorris, line 125: can you await the delete here? or is it synchronous?

sdk/js/src/DefaultCacheStore.ts
elliotCamblor: its not, I mostly did this because having an `init` function is more readable for the `IndexedDBStrategy` implementation and I wanted both classes to have the same interface. I can change this though

sdk/js/src/Client.ts
elliotCamblor: yeah, agreed. I wanted to avoid changing it for the sake of minimizing risk, though I don't love having to mix .then with async/await

sdk/js/src/DefaultCacheStore.ts
elliotCamblor: store.get and store.delete are synchronous. The parent function only needs to be async for the `await this.ready`

kaushalkapasi (approved):

sdk/js/src/Client.ts
```diff
@@ -235,6 +231,10 @@ export class DevCycleClient<
             }
         }
 
+        this.resolveOnInitialized(this)
```
ajwootto, line 234: why move where this is happening? before it only happened if the config fetching process didnt throw

sdk/js/src/Client.ts
```diff
@@ -206,15 +206,11 @@ export class DevCycleClient<
...
-            this.logger.info('Client initialized')
         } catch (err) {
             this.initializeOnConfigFailure(this.user, err)
             return this
         }
 
-        this.eventEmitter.emitInitialized(true)
-
         if (this.user.isAnonymous) {
             this.store.saveAnonUserId(this.user.user_id)
```
ajwootto, line 215: if this is async and being ignored it should have `void ` before it

sdk/js/src/Client.ts
```diff
@@ -723,14 +722,13 @@ export class DevCycleClient<
...
                     this.sdkKey,
                     this.logger,
                     this.options,
-                ).then((res) =>
-                    this.logger.info(`Saved response entity! ${res}`),
                 )
+                this.logger.info(`Saved response entity! ${res}`)
             }
 
             this.userSaved = true
```
ajwootto, line 734: what are the implications of "this.userSaved" not being synchronously set to true as soon as `setUser` is called, like it was before?

elliotCamblor: great question, ill look into that

sdk/js/tsconfig.spec.json
```diff
@@ -4,11 +4,5 @@
...
         "types": ["jest", "node"]
     },
-    "include": [
-        "**/*.test.ts",
-        "**/*.spec.ts",
-        "**/*.d.ts",
-        "jest.config.ts",
-        "**/__mocks__/*.ts"
-    ]
+    "include": ["**/*.test.ts", "**/*.spec.ts", "**/*.d.ts", "jest.config.ts"]
```
ajwootto, line 7: why remove the mocks entry?

sdk/js/src/DefaultCacheStore.ts
```diff
@@ -1,26 +1,41 @@
...
+    constructor(isTesting = false) {
+        super()
+        this.isTesting = isTesting
+        this.init()
     }
 
-    save(storeKey: string, data: unknown): void {
-        this.store?.setItem(storeKey, JSON.stringify(data))
+    async init(): Promise<void> {
+        this.store = this.isTesting ? stubbedLocalStorage : window.localStorage
```
ajwootto, line 25: does this need to be in runtime code? seems like it should be mocked by the tests?

elliotCamblor: not sure, this is how it was done before

elliotCamblor: will see if its easy enough to change

sdk/js/src/ConfigRequestConsolidator.ts
```diff
@@ -68,8 +68,8 @@ export class ConfigRequestConsolidator {
                     // don't resolve anything and just make another request while keeping all the previous resolvers
                     this.resolvers.push(...resolvers)
                 } else {
-                    resolvers.forEach(({ resolve }) => resolve(result))
                     this.handleConfigReceivedFunction(result, this.nextUser)
+                    resolvers.forEach(({ resolve }) => resolve(result))
```
ajwootto, line 72: whats the ordering change here doing again?

elliotCamblor: it was mission critical when I had made the `handleConfigReceivedFunction` async, because it was resolving the promise before all the async operations were complete. Now that were not awaiting the db operations it could go either way. I think it makes more sense to resolve the promise after handling the config being received, but I can also revert it to how it was

ajwootto: probably true

sdk/js/src/Client.ts
elliotCamblor: Looks like its only used to prevent this function from being called multiple times unnecessarily. Dont think theres any real issue here

sdk/js/tsconfig.spec.json
elliotCamblor: It was complaining about the fetch mock I think. MIght not be necessary anymore, will double check

ajwootto: removing it would likely give a type error if trying to use `jest` anywhere in a mocks file

sdk/js/src/DefaultCacheStore.ts
elliotCamblor: is it alright if I make this a separate ticket? Since it was already being done this way I'd rather not mess with it in an already risky pr

elliotCamblor: Made it into a ticket in our triage, Ill try and pick it up during my oncall shift next week

sdk/js/src/Client.ts
```diff
@@ -442,7 +441,7 @@ export class DevCycleClient<
             }
             const config = await this.requestConsolidator.queue(updatedUser)
             if (user.isAnonymous || !user.user_id) {
```
ajwootto, line 443: hmm does this logic actually make sense? why is it checking user and then saving updatedUser?

elliotCamblor: I think that's intentional. It's checking whether the user had an id before being updated (which would add an anon user_id if there isnt one already). If it already had a user_id before being updated, then theres no need to save the anon user_id to storage because theres already one in there

ajwootto: I think it's actually checking if the new user (passed in as `user`) is anonymous, and if so it's then saving the anon user id for this new user. But it's using `updatedUser` because the DVCPopulatedUser class is what generates the anon user id if it's not specified.

ajwootto (approved):

PR #742, opened 2024-02-28
feat: Pass-through rollouts

elliotCamblor:
Change rollout evaluation to be at the same time as segmentation results so that users that don't pass rollout have the opportunity to get put into further targets

The test works by adding a new `All Users` audience that is the last target, so a bunch of tests have updated featureVars to account for the fact that some test users are now being segmented into a feature that they werent before.

TODO make this functionality a project level setting with a split code path (to preserve existing behaviour) rather than just removing the old functionality


jonathannorris:
@elliotCamblor can we close this now?

PR #747, opened 2024-03-08
chore: create workflow to update OpenFeature SDKs

jonathannorris:


jonathannorris:
* **#747** <a href="https://app.graphite.dev/github/pr/DevCycleHQ/js-sdks/747?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

.github/workflows/update-of-sdk.yml
```diff
@@ -0,0 +1,78 @@
...
+      - name: Set branch name
+        shell: bash
+        run: echo "BRANCH_NAME=update-of-sdk" >> $GITHUB_ENV
+
+      - name: Update OpenFeature SDKs to latest
+        run: yarn add @openfeature/core @openfeature/server-sdk @openfeature/web-sdk
+
+      - name: Update @openfeature/server-sdk peerDependencies in sdk/nodejs/package.json
+        run: |
+          cd sdk/nodejs
```
JamieSinn, line 34: You should use this instead https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsworking-directory

.github/workflows/update-of-sdk.yml
```diff
@@ -0,0 +1,78 @@
...
+      - name: Update @openfeature/web-sdk peerDependencies in sdk/openfeature-web-provider/package.json
+        run: |
+          cd sdk/openfeature-web-provider
+          yarn add --peer @openfeature/web-sdk
+          cd ../..
+
+      - name: Check for changes to package.json
+        run: |
+          IS_UPDATED=$(git diff --name-only origin/main package.json sdk/nodejs/package.json sdk/openfeature-web-provider/package.json | wc -l)
+          if [ "$IS_UPDATED" -gt "0" ]; then
```
JamieSinn, line 47: You're comparing strings here, you need to remove the quotes

.github/workflows/update-of-sdk.yml
```diff
@@ -0,0 +1,78 @@
...
+        shell: bash
+        run: |
+          git config --global user.email "github-tracker-bot@taplytics.com"
+          git config --global user.name "DevCycle Automation"
+
+      - name: Set branch name
+        shell: bash
+        run: echo "BRANCH_NAME=update-of-sdk" >> $GITHUB_ENV
+
+      - name: Update OpenFeature SDKs to latest
```
JamieSinn, line 29: Updating to latest isn't always what we want, can you make this a param to be either a specific version or latest?

JamieSinn (changes requested):

.github/workflows/update-of-sdk.yml
jonathannorris: I haven't really seen a situation yet where we don't want to be on latest, going to keep this simple for now.

.github/workflows/update-of-sdk.yml
jonathannorris: ChatGPT decided to throw some shade at you:

The feedback you received points towards a common misunderstanding in shell scripting, especially regarding how comparisons work in bash. In bash, when you use `-gt` (greater than) for numeric comparison, it actually doesn't matter whether the variable is quoted or not; the shell will evaluate the expression correctly in both cases. The quotes are primarily used to ensure that the variable is treated as a single string, which is useful when your variable might contain spaces or special characters. However, in numeric comparisons like the one you're doing, where the variable is expected to contain a number, quotes don't impact the evaluation negatively.

Here's the line in question:

```bash
if [ "$IS_UPDATED" -gt "0" ]; then
```

In this context, `"$IS_UPDATED"` is perfectly fine because bash will correctly interpret the quoted variable as a number for the comparison. The advice to remove quotes might stem from a misunderstanding or a preference for style, but in bash, both of the following conditions are valid and equivalent for numeric comparisons:

```bash
if [ "$IS_UPDATED" -gt "0" ]; then
```

```bash
if [ $IS_UPDATED -gt 0 ]; then
```

If you're facing issues with this script segment not behaving as expected, the problem likely lies elsewhere. For example, ensure that the `git diff --name-only` command and subsequent pipe to `wc -l` are correctly calculating changes, or that the variable `IS_UPDATED` is being set and echoed to `$GITHUB_ENV` correctly.

To clarify, the use of quotes in bash scripting is generally a good practice for string variables to prevent word splitting and globbing. In numeric comparisons, bash handles variables inside quotes correctly, interpreting them as numbers.

.github/workflows/update-of-sdk.yml
```diff
@@ -0,0 +1,78 @@
...
+          cd ../..
+
+      - name: Update @openfeature/web-sdk peerDependencies in sdk/openfeature-web-provider/package.json
+        run: |
+          cd sdk/openfeature-web-provider
+          yarn add --peer @openfeature/web-sdk
+          cd ../..
+
+      - name: Check for changes to package.json
+        run: |
```
JamieSinn, line 41:
```suggestion
        shell: bash
        run: |
```

JamieSinn (approved):

PR #752, opened 2024-03-11
feat: Setup Nest.js SDK

laurawarr:
## Summary
The Nest.js SDK wraps the Node SDK, exposing all the existing methods as well as:
- `DevCycleModule` provides a `DevCycleClient` instance
- `@VariableValue` a parameter decorator for populating a variable value
- `@RequireVariableValue` a request interceptor that guards endpoints if the user is not served the specified values

## Setup
Basic usage would require importing the `DevCycleModule` globally. This requires the SDK key, and optionally may include SDK options.
```
DevCycleModule.forRoot({
    key: "<DVC_SDK_KEY>",
    options: {} // DVC Options
    userFactory: (context) => {
       // returns DVC User Object
    },
    isGlobal: <bool> // defaults to true
}),
 
// Alternative using forRootAsync
 DevCycleModule.forRootAsync({
    useFactory: (config: ConfigService) => ({
        key: config.get('DVC_SDK_KEY') || '',
        options: {}
    }),
    inject: [ConfigService]
}),
```
With this change the `DevCycleClient` can be injected into other controllers and services. The user returned from the user factory will be used when calling the DVC client from the decorators.

TODO: add tests (in this PR), create documentation (later PR)


sdk/nestjs/package.json
```diff
@@ -0,0 +1,19 @@
+{
+    "name": "@devcycle/nestjs-server-sdk",
+    "version": "0.0.1",
+    "description": "The DevCycle NestJS Server SDK used for feature management.",
+    "license": "MIT",
+    "dependencies": {
+        "@devcycle/nodejs-server-sdk": "^1.18.3",
+        "@devcycle/types": "^1.9.0",
+        "@nestjs/common": "^10.3.3",
```
ajwootto, line 9: these need to be peer dependencies

sdk/nestjs/package.json
```diff
@@ -0,0 +1,19 @@
+{
+    "name": "@devcycle/nestjs-server-sdk",
+    "version": "0.0.1",
+    "description": "The DevCycle NestJS Server SDK used for feature management.",
+    "license": "MIT",
+    "dependencies": {
+        "@devcycle/nodejs-server-sdk": "^1.18.3",
+        "@devcycle/types": "^1.9.0",
```
ajwootto, line 8: all these errors are from the Nx dependency checker lint rule. It basically doesn't work properly most of the time. If you can figure out why it thinks these dependencies aren't used then that would be best, but otherwise you can ignore specific packages from being checked in the `.eslintrc` config for this sdk

sdk/nestjs/project.json
```diff
@@ -0,0 +1,68 @@
...
+                "jestConfig": "sdk/nestjs/jest.config.ts",
+                "passWithNoTests": false,
+                "codeCoverage": true
+            },
+            "dependsOn": [
+                {
+                    "target": "build"
+                },
+                {
+                    "target": "build:json",
```
ajwootto, line 36: this shouldnt be here

ajwootto: at least I dont think its necessary

sdk/nestjs/project.json
```diff
@@ -0,0 +1,68 @@
...
+            "executor": "@nx/js:tsc",
+            "outputs": ["{options.outputPath}"],
+            "options": {
+                "outputPath": "dist/sdk/nestjs",
+                "tsConfig": "sdk/nestjs/tsconfig.lib.json",
+                "packageJson": "sdk/nestjs/package.json",
+                "main": "sdk/nestjs/src/index.ts",
+                "assets": ["sdk/nestjs/*.md"]
+            },
+            "dependsOn": [
```
ajwootto, line 51: can check to confirm but im pretty sure you dont need this because all build targets depend on dependency build targets by default

sdk/nestjs/src/DevCycleModule/DevCycleModule.ts
```diff
@@ -0,0 +1,41 @@
...
+import { APP_INTERCEPTOR } from '@nestjs/core'
+import { initializeDevCycle, DevCycleClient, DevCycleOptions } from '@devcycle/nodejs-server-sdk'
+import { RequestInterceptor } from './RequestInterceptor'
+
+@Global()
+@Module({
+    exports: [DevCycleClient],
+})
+export class DevCycleModule {
+    static forRoot(sdkKey: string, options?: DevCycleOptions): DynamicModule {
```
ajwootto, line 11: you might wanna add a forRootAsync so people can inject the SDK key from somewhere else

sdk/nestjs/src/DevCycleModule/DevCycleModule.ts
```diff
@@ -0,0 +1,41 @@
+import { Module, DynamicModule, Global } from '@nestjs/common'
+import { APP_INTERCEPTOR } from '@nestjs/core'
+import { initializeDevCycle, DevCycleClient, DevCycleOptions } from '@devcycle/nodejs-server-sdk'
+import { RequestInterceptor } from './RequestInterceptor'
+
+@Global()
```
ajwootto, line 6: is it ok for a lib to force itself to be global like this? or is it more expected that the end user can decide? I believe I've seen other libs take a "global" setting in the `forRoot` method that decides it

sdk/nestjs/package.json
ajwootto: yeah i was talking about the nest ones specifically

laurawarr: ah, unless you were referring to "these" as the nestjs packages below this one, rather than the others shown in this preview 👀

sdk/nestjs/src/RequireVariableValue.ts
```diff
@@ -0,0 +1,69 @@
...
+            const request = context.switchToHttp().getRequest()
+            const { dvc_user } = request
+
+            if (!dvc_user) {
+                throw new Error(
+                    'User context not found. Make sure to add the DevCycleUserGuard.',
+                )
+            }
+
+            const servedVariables = this.dvcClient.allVariables(dvc_user)
```
ajwootto, line 40: could we use a series of variable calls here so evaluations get tracked?

sdk/nestjs/src/RequireVariableValue.ts
```diff
@@ -0,0 +1,69 @@
...
+                requiredVariableValues,
+            )) {
+                const { value: servedValue } = servedVariables[key] ?? {}
+                const isDefaulted = servedValue === undefined
+
+                // Manually track the evaluation of the variable
+                this.dvcClient.track(dvc_user, {
+                    type: isDefaulted
+                        ? 'variableDefaulted'
+                        : 'variableEvaluated',
```
ajwootto, line 52: these event types wouldn't be correct since the server sdks track using the `aggVariableEvaluated` event to prevent having per-user events getting sent

sdk/nestjs/src/DevCycleUserGuard.ts
```diff
@@ -0,0 +1,14 @@
+import { Injectable, CanActivate, ExecutionContext } from '@nestjs/common'
+import { DevCycleUser } from '@devcycle/nodejs-server-sdk'
+
+@Injectable()
+export abstract class DevCycleUserGuard implements CanActivate {
```
ajwootto, line 5: isn't this more like middleware than a guard?
https://docs.nestjs.com/middleware

it's not actually guarding anything its only function is to add stuff to the request

sdk/nestjs/src/DevCycleModule/DevCycleModule.ts
```diff
@@ -0,0 +1,54 @@
...
+export class DevCycleModule extends ConfigurableModuleClass {}
+
+async function createClient({ key, options }: DevCycleModuleOptions): Promise<DevCycleClient> {
+    if (!key.trim()) {
+        throw new Error('Missing required DevCycle SDK key')
+    }
+
+    return await initializeDevCycle(key, {
+        ...options,
+        enableCloudBucketing: false,
```
ajwootto, line 52: if this can't be configured it should probably be omitted from the options type

sdk/nestjs/src/DevCycleUserGuard.ts
laurawarr: Yep, although the main challenge is that would mean that any logic that's used to parse/define the user object would also need to be handled in middleware. In our case we use a guard to parse the jwt token on authenticated routes and I suspect that would often be the case.

I'd been back and forth about whether to implement this and the `RequireVariableValue` decorator as guards or interceptors, but since RequireVariableValue determines whether an endpoint can be accessed it makes the most sense as a guard.

sdk/nestjs/src/RequireVariableValue.ts
laurawarr: I originally tried that, but we would need to pass a default value, which doesn't really make sense in this context

sdk/nestjs/src/DevCycleUserGuard.ts
ajwootto: Would it be more of an interceptor then? It sounds like those run after guards.

I think RequireVariableValue is definitely a guard though, so I guess it needs access to the user by the time it runs..

ajwootto: what about passing the user getter into the RequireVariableValue or possibly registering it beforehand in the module?

sdk/nestjs/src/VariableValue.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export const VariableValue = createParamDecorator(
+    (data: VariableValueData, ctx: ExecutionContext) => {
+        const req = ctx.switchToHttp().getRequest()
+        if (!req.dvc_user) {
+            throw new Error(
+                'User context not found. Make sure to add the DevCycleUserGuard.',
+            )
+        }
+        return req.dvc_client.variableValue(
```
ajwootto, line 17: should this check for the existence of dvc_client too?

sdk/nestjs/src/DevCycleModule/RequestInterceptor.ts
```diff
@@ -0,0 +1,23 @@
...
+@Injectable()
+export class RequestInterceptor implements NestInterceptor {
+    constructor(@Inject(DevCycleClient) private client: DevCycleClient) {}
+
+    intercept(
+        context: ExecutionContext,
+        next: CallHandler,
+    ): CallHandler['handle'] {
+        const req = context.switchToHttp().getRequest()
+        req.dvc_client = this.client
```
ajwootto, line 19: I don't know if it's a problem or not to be adding our own stuff to the request object, but maybe another option is to use Continuous local storage to access the client anywhere in the request context?
https://papooch.github.io/nestjs-cls/introduction/quick-start

sdk/nestjs/package.json
```diff
@@ -0,0 +1,18 @@
...
+        "@devcycle/nodejs-server-sdk": "^1.18.3"
+    },
+    "peerDependencies": {
+        "@nestjs/common": "^10.3.3",
+        "@nestjs/core": "^10.3.3"
+    },
+    "engines": {
+        "node": ">=16.0.0"
+    },
+    "main": "src/index.ts",
```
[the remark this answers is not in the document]
laurawarr: fair, although it looks like almost all of our package.json files are pointing to a ts file for main 👀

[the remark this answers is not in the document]
ajwootto: 👀

sdk/nestjs/src/DevCycleModule/RequestInterceptor.ts
```diff
@@ -0,0 +1,32 @@
+import { Observable } from 'rxjs'
```
ajwootto, line 1: is rxjs dependency just being added here in order to import a type?

laurawarr: Yeah. The linter is unhappy if the response type of `intercept` isn't defined, and it's unhappy if rxjs is just added as a dev dependency. I supposed I could leave it as a dev dependency and just update linting config

ajwootto: can you just type it as "unknown" ?

ajwootto: or use `typeof TheThingYouAreReturning`

laurawarr: Using unknown doesn't satisfy the NestInterceptor interface, and the thing we're returning is the result of `CallHandler.handle()` which is an rxjs Observable.

It's just the linting rule that wants exported response types to be explicitly defined. But considering typescript can infer the types just fine, and this interceptor is just used internally, we could ignore the rule for this case

ajwootto: yeah just ignore it

ajwootto: did ignoring it not work?

laurawarr: I get a build error that the type can't be implied if I try to remove the return type and ignore this line. So instead I added rxjs as a dev dependency and update the lint config to ignore the dependency error for rxjs

ajwootto: what about `ReturnType<CallHandler['handle']>`

laurawarr: cool, that works!

package.json
```diff
@@ -184,6 +186,7 @@
         "react-native-svg-transformer": "^1.0.0",
         "react-refresh": "^0.10.0",
         "react-test-renderer": "18.2.0",
+        "rxjs": "^7.8.1",
```
ajwootto, line 189: can remove this

sdk/nestjs/src/VariableValue.test.ts
```diff
@@ -0,0 +1,59 @@
...
+    })
+
+    it('returns value from devcycle variableValue', () => {
+        const key = 'myVariable'
+        const defaultValue = 'default value'
+
+        const factory = getParamDecoratorFactory(VariableValue)
+        const value = factory({ key, default: defaultValue }, ctx)
+
+        expect(value).resolves.toBe('variable value')
```
ajwootto, line 57: nitpick but this doesnt really need to be a separate test case

package.json
laurawarr: I forgot this was resolving the `Cannot find module 'rxjs/operators'` error in the RequestInterceptor test

ajwootto (approved):

PR #754, opened 2024-03-12
chore: update GH Action Upload CDN Release task to upload to versioned paths

kaushalkapasi:
- chore: update GH Action Upload CDN Release task to upload to versioned paths
- chore: remove sleuth from release process

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -104,5 +104,9 @@ jobs:
                   yarn nx build:cdn js
 
             - name: Upload CDN Release
-              run: aws s3 sync dist/sdk/cdn/js s3://js.devcycle.com/ --include "devcycle.min.js*" --acl public-read
+              run: |
+                JQ_PATH=".version"
+                SDK_VERSION="$(cat package.json | jq -r ${{ JQ_PATH }})"
+                aws s3 sync dist/sdk/cdn/js s3://js.devcycle.com/latest/ --include "devcycle.min.js*" --acl public-read
+                aws s3 sync dist/sdk/cdn/js s3://js.devcycle.com/${{ SDK_VERSION }}/ --include "devcycle.min.js*" --acl public-read
               if: inputs.draft != true && github.ref == 'refs/heads/main'
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -101,8 +101,8 @@ jobs:
 
             - name: Upload CDN Release
               run: |
-                JQ_PATH=".version"
-                SDK_VERSION="$(cat sdk/js/package.json | jq -r ${{ JQ_PATH }})"
+                SDK_VERSION="$(cat sdk/js/package.json | jq -r ".version""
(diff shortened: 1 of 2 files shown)


kaushalkapasi:
* **#754** <a href="https://app.graphite.dev/github/pr/DevCycleHQ/js-sdks/754?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

 Join @kaushalkapasi and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

.github/workflows/release.yml
```diff
@@ -88,21 +88,21 @@ jobs:
...
 
             - name: Build UMD bundles
               run: |
                   yarn nx build:cdn js
 
             - name: Upload CDN Release
-              run: aws s3 sync dist/sdk/js s3://js.devcycle.com/ --include "devcycle.min.js*" --acl public-read
+              run: |
+                JQ_PATH=".version"
+                SDK_VERSION="$(cat package.json | jq -r $JQ_PATH)"
```
ajwootto, line 105: what package.json is being read here?

kaushalkapasi: ah.. right now this will read the root `package.json` file, which isn't the right one... let me update to point to the `sdk/js/package.json` file as that will be the right SDK version

.github/workflows/release.yml
```diff
@@ -88,21 +88,21 @@ jobs:
...
               if: inputs.draft == true
 
             - name: Build UMD bundles
               run: |
                   yarn nx build:cdn js
 
             - name: Upload CDN Release
-              run: aws s3 sync dist/sdk/cdn/js s3://js.devcycle.com/ --include "devcycle.min.js*" --acl public-read
+              run: |
+                JQ_PATH=".version"
```
ajwootto, line 104: doesn't really seem like this needs to be assigned to a variable and could just be written directly on the next line

.github/workflows/release.yml
```diff
@@ -88,21 +88,21 @@ jobs:
...
+                SDK_VERSION="$(cat sdk/js/package.json | jq -r ${{ JQ_PATH }})"
+                aws s3 sync dist/sdk/cdn/js s3://js.devcycle.com/latest/ --include "devcycle.min.js*" --acl public-read
```
ajwootto, line 106: hmm wait this changes the path the current CDN file is being uploaded on, that's gonna prevent everyone's current setups from receiving any more updates

ajwootto (changes requested):
comment about changing the upload path

.github/workflows/release.yml
kaushalkapasi: yeah, do we want this to be an explicit change over? that way current users aren't impacted

ajwootto: but why wouldn't we want to keep uploading to the existing path with the latest version?

kaushalkapasi: will make the change to update the existing path as well 👍

PR #755, opened 2024-03-13
feat: cor 2573 unrevert axios replacement

elliotCamblor:
- Removed cross-fetch from js-client-sdk dependencies and moved to devDependencies for testing purposes
- Diverged the RequestUtils and server-requests lib with this change ^


sdk/js/src/RequestUtils.ts
```diff
@@ -0,0 +1,159 @@
...
+
+    return handleResponse(res)
+}
+
+async function getFetch() {
+    if (typeof fetch !== 'undefined') {
+        return fetch
+    }
+
+    return (await import(/* webpackMode: "eager" */ 'cross-fetch')).default
```
ajwootto, line 136: why dynamically import this if forcing webpack to always include it anyway?

also why do we actually need cross-fetch again?

elliotCamblor: Because we have more than just a webpack build,
and I believe it's to import fetch for environments without access to it natively

ajwootto: what environments is the JS SDK used in that dont have access to fetch?

jsalaber (approved):

PR #768, opened 2024-03-18
fix: update the AbortError to indicate a network connection timeout

elliotCamblor:


package.json
```diff
@@ -158,7 +158,7 @@
         "jest": "^29.7.0",
         "jest-bench": "^29.4.1",
         "jest-circus": "~27.2.3",
-        "jest-cli": "~29.7.0",
+        "jest-cli": "~28.1.3",
```
jsalaber, line 161: any reason to downgrade this?

elliotCamblor: it said it wasnt found when I was running yarn. Dependabot updated it so I reverted to be safe

elliotCamblor: ill see if I can un-downgrade it

sdk/js/src/RequestUtils.ts
```diff
@@ -61,15 +61,23 @@ export async function getWithTimeout(
...
+        }, timeout)
+        const response = await get(url, {
+            ...requestConfig,
+            signal: controller.signal,
+        })
+        clearTimeout(id)
+        return response
+    } catch (e) {
+        if (controller?.signal?.aborted) {
+            throw new Error('Network connection timed out.')
```
jsalaber, line 76: possible to make a test for this?

elliotCamblor: I don't think it would be a very effective test, because the controller.signal and controller.abort stuff is part of the fetch api which Im mocking out

jsalaber (approved):

sdk/js/src/RequestUtils.ts
elliotCamblor: Ill see if I can add an artificial 5 second delay to the fetch mock and see what happens

PR #781, opened 2024-03-26
chore: remove nrwl jest package

ajwootto:
- removes the @nrwl/jest package that was added during e2e test work. It's set to the wrong (newest) version and also packages have been renamed to @nx/*, and we already use `@nx/jest`


suthar26 (approved):
makes sense, didnt clean up properly between iterations

PR #786, opened 2024-03-27
feat: Add mocks to nestjs sdk

laurawarr:
The SDK can be mocked using `jest.mock('@devcycle/nestjs-server-sdk')`

Variable values can be mocked using the `mockVariables` method
```
  const dvcClient = moduleRef.get<DevCycleClient>(
      DevCycleClient,
  ) as MockDevCycleClient
  dvcClient.mockVariables({ 'user-permissions': true })
```

### DIFF
--- a/sdk/nestjs/tsconfig.lib.json
+++ b/sdk/nestjs/tsconfig.lib.json
@@ -8,7 +8,6 @@
     "exclude": [
         "**/*.spec.ts",
         "**/*.test.ts",
-        "**/__mocks__/**/*.ts",
         "**/__tests__/**/*.ts",
         "jest.config.ts"
     ],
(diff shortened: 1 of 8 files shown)


sdk/nestjs/tsconfig.lib.json
```diff
@@ -8,9 +8,8 @@
     "exclude": [
         "**/*.spec.ts",
         "**/*.test.ts",
-        "**/__mocks__/**/*.ts",
         "**/__tests__/**/*.ts",
         "jest.config.ts"
     ],
-    "include": ["**/*.ts"]
+    "include": ["**/*.ts", "mocks.ts"]
```
jsalaber, line 14: does `**/*.ts` not already satisfy the path for `mocks.ts`?

ajwootto (approved):

sdk/nestjs/tsconfig.lib.json
laurawarr: yep, you're right — I think when I originally added this the file was `mocks.js`

PR #788, opened 2024-03-28
feat: Add DevCycleService to Nest SDK

laurawarr:
Add DevCycleService with methods that use the user object returned from the `userFactory`

### DIFF
--- a/sdk/nestjs/src/DevCycleModule/RequestWithData.ts
+++ b/sdk/nestjs/src/DevCycleModule/RequestWithData.ts
@@ -1,0 +1,14 @@
+import { ExecutionContext } from '@nestjs/common'
+import { DevCycleClient, DevCycleUser } from '@devcycle/nodejs-server-sdk'
+
+export interface RequestWithData {
+    [key: string]: any
+    get dvc_client(): DevCycleClient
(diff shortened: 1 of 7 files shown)


sdk/nestjs/src/DevCycleModule/RequestWithData.ts
```diff
@@ -0,0 +1,24 @@
+import { Request } from 'express'
```
ajwootto, line 1: what if they aren't using express as the HTTP adapter? 

also there should be a way to get this type without importing it from the express package via inference from the Nest types. We shouldn't add a direct dependency on express with this package either

laurawarr: Nest's docs suggest importing the request types from express, we could instead leave the request type as any object with the dvc_client and dvc_user properties defined

ajwootto: yeah if the type isn't meant to be used externally to this package then just type it as any, better that than adding a dependency that needs to be installed just for internal TS typing benefit

ajwootto (approved):

PR #792, opened 2024-04-02
fix: Update nest sdk to use nest-cls

laurawarr:
Use nest-cls to manage instance of dvc client and user object rather than request object.

With the current implementation, when using the dvc service within the event emitter context the injected request object is actually the event, which causes an error. Using cls instead works as expected


sdk/nestjs/src/DevCycleModule/DevCycleService.ts
```diff
@@ -25,8 +25,8 @@ export class DevCycleService {
     variableValue<T extends DVCVariableValue>(
         key: string,
         defaultValue: T,
-    ): DVCVariableValue {
-        return this.devcycleClient.variableValue(
+    ): VariableTypeAlias<T> {
```
ajwootto, line 28: could also just type this as the ReturnType of the client's variableValue method

ajwootto: or not define it assuming that doesn't cause a linting error or something

sdk/nestjs/src/VariableValue.ts
```diff
@@ -9,7 +8,7 @@ type VariableValueData = {
 
 export const VariableValue = createParamDecorator(
     (data: VariableValueData, ctx: ExecutionContext) => {
-        const req = getRequestFromContext(ctx)
+        const req = ctx.switchToHttp().getRequest()
 
         if (!req.dvc_client || !req.dvc_user) {
```
ajwootto, line 13: should this be using CLS?

ajwootto: or is it gonna have both CLS and the request properties?

laurawarr: I was thinking it needed both since cls can't be injected here, but it seems like I can use ClsServiceManager to get the cls service

sdk/nestjs/src/DevCycleModule/DevCycleService.ts
laurawarr: I'm fairly sure leaving it undefined will cause a lint error — updated to use the return type

sdk/nestjs/.eslintrc.json
```diff
@@ -14,7 +14,7 @@
                 "@nx/dependency-checks": [
                     "error",
                     {
-                        "ignoredDependencies": []
+                        "ignoredDependencies": ["nestjs-cls"]
```
ajwootto, line 17: is the linter complaining about this? what's it saying?

sdk/nestjs/src/DevCycleModule/DevCycleService.ts
```diff
@@ -25,8 +24,8 @@ export class DevCycleService {
     variableValue<T extends DVCVariableValue>(
         key: string,
         defaultValue: T,
-    ): DVCVariableValue {
-        return this.devcycleClient.variableValue(
+    ): ReturnType<DevCycleClient['variableValue']> {
```
ajwootto, line 27: is this correct without the generic being specified?

sdk/nestjs/.eslintrc.json
laurawarr: it's complaining that the package is referenced in package.json but it's not used by the project

sdk/nestjs/src/DevCycleModule/DevCycleService.ts
ajwootto: (the returntype)

laurawarr: Probably not... this is surprisingly annoying to get the correct return type here

ajwootto: personally I would just disable that linting rule and let it infer here, assuming it compiles alright. Its usually better to not explicitly define it anyway since you run the risk of changing the type in subtle ways compared to what it should be

laurawarr: Using the ReturnType with the generic would be preferable, but I haven't found a way to do that

PR #838, opened 2024-05-03
feat: add SSEConnection shared lib

jonathannorris:

PR #840, opened 2024-05-03
feat: implement SSE connections in NodeJS config manager

jonathannorris:


jonathannorris:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/DevCycleHQ/js-sdks/840?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#840** <a href="https://app.graphite.dev/github/pr/DevCycleHQ/js-sdks/840?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#838** <a href="https://app.graphite.dev/github/pr/DevCycleHQ/js-sdks/838?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

lib/shared/config-manager/src/index.ts
```diff
@@ -67,26 +86,126 @@ export class EnvironmentConfigManager {
...
+    stopWatchingForConfigChanges(): void {
+        this.configFetchState = ConfigFetchState.DISABLED
+        this.stopPolling()
+        this.stopSSE()
+    }
+
+    private startSSE(): void {
+        if (this.configFetchState === ConfigFetchState.POLLING) {
+            this.stopPolling()
+        }
```
JamieSinn, line 117: you should only disable polling after SSE has successfully connected

lib/shared/config-manager/src/index.ts
```diff
@@ -67,26 +86,126 @@ export class EnvironmentConfigManager {
...
+                    //     true,
+                    //     messageData.lastModified,
+                    //     messageData.etag,
+                    // ).catch((e) => {
+                    //     this.logger.warn(`Failed to refetch config ${e}`)
+                    // })
+                }
+            }
+        } catch (e) {
+            this.logger.warn(`Streaming Connection: Unparseable message ${e}`)
```
JamieSinn, line 174: should we include the raw message too?

lib/shared/config-manager/src/index.ts
```diff
@@ -67,26 +86,126 @@ export class EnvironmentConfigManager {
...
+        this.stopSSE()
+    }
+
+    private startSSE(): void {
+        if (this.configFetchState === ConfigFetchState.POLLING) {
+            this.stopPolling()
+        }
+
+        if (!this.configSSE) {
+            this.logger.warn('No SSE configuration found, switching to polling')
```
ajwootto, line 120: what's this warning supposed to indicate to the user? and what would they do about it

lib/shared/config-manager/src/index.ts
```diff
@@ -184,10 +291,32 @@ export class EnvironmentConfigManager {
...
         }
     }
+
+    private handleSSEConfig(projectConfig: string) {
+        if (this.enableRealtimeUpdates) {
+            const configBody = JSON.parse(projectConfig) as ConfigBody<string>
+            const originalConfigSSE = this.configSSE
+            this.configSSE = configBody.sse
+
+            // Reconnect SSE if not first config fetch, and the SSE config has changed
```
jonathannorris, line 307: Does this reconnecting logic make sense?

lib/shared/config-manager/src/index.ts
```diff
@@ -1,13 +1,16 @@
...
+import { SSEConnection } from '@devcycle/sse-connection'
 
 type ConfigPollingOptions = {
     configPollingIntervalMS?: number
+    sseConfigPollingIntervalMS?: number
     configPollingTimeoutMS?: number
     configCDNURI?: string
     cdnURI?: string
     clientMode?: boolean
+    betaEnableRealTimeUpdates?: boolean
```
JamieSinn, line 13: i have this as `EnableBetaRealtimeUpdates` in go - do we want to align them?

lib/shared/config-manager/src/index.ts
```diff
@@ -63,26 +84,115 @@ export class EnvironmentConfigManager {
...
+                this.logger.debug('SSE connection opened')
+                // Set config polling interval to 10 minutes
+                this.startPolling(this.sseConfigPollingIntervalMS)
+            },
+            onConnectionError: () => {
+                this.logger.debug('SSE connection error, switching to polling')
+                // reset polling interval to default
+                this.startPolling(this.configPollingIntervalMS)
+                this.stopSSE()
+            },
```
JamieSinn, line 121: Is there a risk that the polling does not start?

lib/shared/config-manager/src/index.ts
JamieSinn: How can SSE be "started" before there is a config downloaded? Might make sense to track the state similar to go?

lib/shared/sse-connection/src/lib/SSEConnection.ts
```diff
@@ -20,15 +27,17 @@ export class SSEConnection {
...
         this.connection = new EventSource(this.url, { withCredentials: true })
         this.connection.onmessage = (event) => {
-            this.onMessage(event.data)
+            this.callbacks.onMessage(event.data)
         }
         this.connection.onerror = () => {
             this.logger.warn(
                 'StreamingConnection warning. Connection failed to establish.',
             )
+            this.callbacks.onConnectionError()
```
JamieSinn, line 36: Do you have a handler for when the connection is interrupted after it's already connected?

jonathannorris: there is only a single callback on the EventSource library, and it doesn't distinguish between them: 
image

lib/shared/config-manager/src/index.ts
jonathannorris: this function gets called during the processing of the config response, including the first config download.

lib/shared/config-manager/src/index.ts
jonathannorris: nope

lib/shared/sse-connection/src/lib/SSEConnection.ts
JamieSinn: I think you might be flapping a lot if you're not doing some sort of error handling checks to make sure the connection is actually dead

examples/nodejs-local/src/main.ts
```diff
@@ -10,7 +10,8 @@ let devcycleClient: DevCycleClient
 
 async function startDevCycle() {
     devcycleClient = await initializeDevCycle(DEVCYCLE_SERVER_SDK_KEY, {
-        logLevel: 'info',
+        logLevel: 'debug',
+        enableBetaRealTimeUpdates: true,
```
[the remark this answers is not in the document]
ajwootto: the example is for local development so whatever makes it easier for developing the SDK locally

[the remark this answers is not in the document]
jonathannorris: this is just from my testing, will cleanup.

lib/shared/config-manager/__tests__/environmentConfigManager.spec.ts
```diff
@@ -223,4 +261,289 @@ describe('EnvironmentConfigManager Unit Tests', () => {
...
+                            lastModifiedDate.getTime() + 1000,
+                        ).toISOString(),
+                    }),
+                }),
+            })
+            jest.advanceTimersByTime(1005)
+            expect(getEnvironmentConfig_mock).toBeCalledTimes(2)
+
+            envConfig.cleanup()
+        })
```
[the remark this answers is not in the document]
jonathannorris: well these are pure unit tests of the code with a bunch of stuff mocked, we should also have test-harness tests. But these helped me catch a bunch of issues with my implementation.

JamieSinn (approved):

lib/shared/config-manager/src/index.ts
```diff
@@ -63,28 +84,121 @@ export class EnvironmentConfigManager {
...
+        })
+    }
+
+    private onSSEMessage(message: string): void {
+        this.logger.debug(`SSE message: ${message}`)
+        try {
+            const parsedMessage = JSON.parse(message as string)
```
ajwootto, line 128: message is already a string

lib/shared/config-manager/src/index.ts
```diff
@@ -63,28 +84,121 @@ export class EnvironmentConfigManager {
...
+            onOpen: () => {
+                this.logger.debug('SSE connection opened')
+                // Set config polling interval to 10 minutes
+                this.startPolling(this.sseConfigPollingIntervalMS)
+            },
+            onConnectionError: () => {
+                this.logger.debug('SSE connection error, switching to polling')
+                // reset polling interval to default
+                this.startPolling(this.configPollingIntervalMS)
+                this.stopSSE()
```
ajwootto, line 120: why does it need to stop and kill the SSE connection when theres an error? can't it just reconnect?

lib/shared/config-manager/src/index.ts
```diff
@@ -162,16 +277,24 @@ export class EnvironmentConfigManager {
             )
             return
         } else if (res?.status === 200 && projectConfig) {
+            const lastModifiedHeader = res?.headers.get('last-modified')
+            if (this.isLastModifiedHeaderOld(lastModifiedHeader)) {
+                this.logger.warn(
+                    'Skipping saving config, existing last modified date is newer.',
```
ajwootto, line 283: is this important enough to be a warning? doesnt this just indicate things are working as expected?

lib/shared/config-manager/src/index.ts
```diff
@@ -186,10 +309,48 @@ export class EnvironmentConfigManager {
...
+
+    private handleSSEConfig(projectConfig: string) {
+        if (this.enableRealtimeUpdates) {
+            const configBody = JSON.parse(projectConfig) as ConfigBody<string>
+            const originalConfigSSE = this.configSSE
+            this.configSSE = configBody.sse
+
+            // Reconnect SSE if not first config fetch, and the SSE config has changed
+            if (
+                this.hasConfig &&
```
ajwootto, line 342: when could hasConfig be false when this is called?

ajwootto: oh is this to prevent it from happening on the first fetch?

lib/shared/config-manager/src/index.ts
```diff
@@ -63,28 +84,121 @@ export class EnvironmentConfigManager {
...
-                }
-                this.intervalTimeout = this.setInterval(async () => {
-                    try {
-                        await this._fetchConfig()
-                    } catch (ex) {
-                        this.logger.error((ex as Error).message)
-                    }
-                }, this.pollingIntervalMS)
+                this.startPolling(this.configPollingIntervalMS)
+                this.startSSE()
```
ajwootto, line 88: why do you need a special case of starting SSE here rather than just letting the existing config fetch handler do it?

lib/shared/config-manager/src/index.ts
```diff
@@ -63,28 +84,121 @@ export class EnvironmentConfigManager {
...
+                )
+                return
+            }
+
+            this._fetchConfig()
+                .then(() => {
+                    this.logger.debug('Config re-fetched from SSE message')
+                })
+                .catch((e) => {
+                    this.logger.warn(
```
ajwootto, line 153: is it good enough to just log that it failed and not keep retrying? before a failure wasnt as big of a deal because it would retry at the polling interval which was every minute or so. 10 minutes is a pretty long time to not receive an update.

lib/shared/config-manager/src/index.ts
jonathannorris: good question, I need to double-check that.

lib/shared/config-manager/src/index.ts
jonathannorris: my understanding of the SSE library is that it already re-connects automatically, and this get's called if it fails to re-connect. So in that case we should just close the SSE connection and go back to polling. But will double check that.

lib/shared/config-manager/src/index.ts
jonathannorris: There is retrying in the fetch code, so if this fails we probably should go back to the shorter polling interval. But that also probably means we should then disconnect SSE, because I'm trying to keep this code somewhat simple in its state, either SSE is connected and the polling interval is long, or its disconnected and doing regular polling.

jonathannorris: Also just noticed we don't really have anything that's verifying that the config returned in this fetch call is equal to the one from the message or newer...

lib/shared/config-manager/src/index.ts
jonathannorris: yea

lib/shared/config-manager/src/index.ts
jonathannorris: I think it makes sense to keep starting SSE + Polling here, as it ensures that polling is started the first time before SSE is started, and that the polling code doesn't need to understand if SSE is connected or not. The SSE code will handle changing the polling interval once it has connected.

lib/shared/config-manager/src/index.ts
jonathannorris: it looks like it will try and reconnect every 1 sec: https://github.com/EventSource/eventsource/blob/db562b82fffb58b77cd6bb817a835db5c97a2398/lib/eventsource.js#L55

ajwootto: and when does it call the error handler?

jonathannorris: check the `_emit('error', new Event('error')` messages in that code. I believe those are all the calls to `onerror`. Its not really clear to me if it ever not retries the connection. Maybe we need to close the connection based on the status codes?

lib/shared/config-manager/src/index.ts
jonathannorris: ok based on chatting with ChatGPT I think we should switch to setting:
```
We use If-Match and If-Unmodified-Since headers to ensure that the server returns the file only if it matches the given ETag and Last-Modified values.
```

which will return a 412 error if it doesn't match: 

```
If the server returns a 412 Precondition Failed status, it means that the resource does not match the provided ETag or Last-Modified values. You can handle this accordingly to fetch the latest resource.
```

lib/shared/config-manager/src/request.ts
```diff
@@ -1,24 +1,78 @@
...
+    let retries = 2
+
+    let retryOn: RequestInitWithRetry['retryOn'] | undefined
+    const sseLastModifiedDate = sseLastModified
+        ? new Date(sseLastModified)
+        : null
+
+    // Retry fetching config if the Last-Modified header is older than
+    // the requiredLastModified from the SSE message
+    if (sseLastModified && isValidDate(sseLastModifiedDate)) {
```
jonathannorris, line 33: Updated retry logic for SSE config requests.

lib/shared/config-manager/src/index.ts
JamieSinn: Do we know for certain that Cloudflare supports this?

jonathannorris: we didn't end up going this approach and went with custom retry logic.

lib/shared/config-manager/src/request.ts
```diff
@@ -1,24 +1,78 @@
...
+        retries = 5
+        retryOn = (attempt, error, response) => {
+            if (attempt >= retries) {
+                return false
+            } else if (response && response?.status === 200) {
+                const lastModifiedHeader =
+                    response?.headers.get('Last-Modified')
```
ajwootto, line 40: response has to exist here

ajwootto (approved):
should fill out the PR description with a list of main changes

PR #841, opened 2024-05-06
feat: updating bucketing passthrough

elliotCamblor:
Bucketing:
- Add another check for rollouts that coincides with the segmentation check. This means that if a user doesnt pass rollout the bucketing logic will progress to the next targeting rule rather than serve no variable values.
- This check will only happen if `disablePassthroughRollouts` in project settings is false.
- Also add a check for the project setting to the original rollout calculation so that the rollout isnt checked a 2nd time.
- Updated the bucketin-AS config body type to have a Settings class for the project field. This class will _only_ parse out the `disablePassthroughRollouts` setting, as no other project settings are used by the bucketing lib.
Tests:
- Add an `all users` target to the end of one of the features. This will ensure that passthrough rollouts will result in the test user being served variable values when they dont meet the first target's rollout.
- This has a knockon effect for some other tests that needed to be updated.
- Also added a test for the original `disablePassthroughRollouts: true` case which should NOT result in the user being put in the all users target


lib/shared/bucketing-assembly-script/assembly/types/configBody.ts
```diff
@@ -13,30 +13,44 @@ import { Feature } from './feature'
...
-        this.key = getStringFromJSON(project, 'key')
-        this.a0_organization = getStringFromJSON(project, 'a0_organization')
-        this.settings = getJSONObjFromJSON(project, 'settings')
+        super();
+        this._id = getStringFromJSON(project, '_id');
+        this.key = getStringFromJSON(project, 'key');
+        this.a0_organization = getStringFromJSON(project, 'a0_organization');
+
+        // Parse settings directly within this class
+        let settings = getJSONObjFromJSON(project, 'settings');
```
jonathannorris, line 28: Maybe `ProjectSettings` should be its own Class and parsed the same way the other sub-objects are?

lib/shared/bucketing-assembly-script/assembly/types/configBody.ts
```diff
@@ -13,30 +13,44 @@ import { Feature } from './feature'
...
-        const json = new JSON.Obj()
-        json.set('_id', this._id)
-        json.set('key', this.key)
-        json.set('a0_organization', this.a0_organization)
-        json.set('settings', this.settings)
+        const json = new JSON.Obj();
+        json.set('_id', this._id);
+        json.set('key', this.key);
+        json.set('a0_organization', this.a0_organization);
+        json.set('disablePassthroughRollouts', this.disablePassthroughRollouts); // Store directly
```
jonathannorris, line 47: I think this might be your issue, previously you were parsing and writing `settings` as a `JSON.Obj`. Now you are parsing `settings` and picking out `disablePassthroughRollouts`, then setting it as a top-level boolean.

elliotCamblor: this was just something I tried in my latest iteration. Before this I was parsing it out into a `settings` subfield. Ill revert back to that state

lib/shared/bucketing-assembly-script/__tests__/bucketing/bucketing.test.ts
```diff
@@ -1309,7 +1446,7 @@ describe('Client Data', () => {
         const c1 = generateBucketedConfig(user)
         expect(c1).toEqual(
             expect.objectContaining({
-                featureVariationMap: {},
+                featureVariationMap: {'614ef6aa473928459060721a': '6153553b8cf4e45e0464268d'},
```
jonathannorris, line 1449: Why are these added here now?

lib/shared/bucketing-assembly-script/assembly/bucketing/bucketing.ts
```diff
@@ -141,13 +141,20 @@ function evaluateSegmentationForFeature(
     // Returns the first target for which the user passes segmentation
     for (let i = 0; i < feature.configuration.targets.length; i++) {
         const target = feature.configuration.targets[i]
+        const boundedHashData = _generateBoundedHashes(user.user_id, target._id)
+        const rolloutHash = boundedHashData.rolloutHash
```
jonathannorris, line 145: Can we move this inside of `if (target.rollout && passthroughRolloutEnabled) {`, so we don't generate the hash if its not enabled.

lib/shared/bucketing-assembly-script/__tests__/bucketing/bucketing.test.ts
elliotCamblor: I mentioned this in the description briefly, but its because I added the all users target to the test data

lib/shared/bucketing-assembly-script/assembly/bucketing/bucketing.ts
elliotCamblor: The bounded hash is still used for bucketing

jonathannorris: In this function? it seems like `rolloutHash` is only ever used here in `_doesUserPassRollout()`

lib/shared/bucketing-assembly-script/assembly/types/configBody.ts
```diff
@@ -8,32 +8,52 @@ import {
...
+        return json.stringify()
+    }
+}
+
 export class PublicProject extends JSON.Value {
-    readonly _id: string
-    readonly key: string
-    readonly a0_organization: string
-    readonly settings: JSON.Obj
+    readonly _id: string;
```
jonathannorris, line 36: Can you remove all the `;` you added?

elliotCamblor: yeah, idk why prettier isnt getting rid of them

lib/shared/bucketing-assembly-script/assembly/bucketing/bucketing.ts
```diff
@@ -141,13 +141,20 @@ function evaluateSegmentationForFeature(
...
+            doesUserPassRollout = _doesUserPassRollout(target.rollout, rolloutHash)
+        }
         if (
             _evaluateOperator(
                 target._audience.filters,
                 config.audiences,
                 user,
                 clientCustomData,
-            )
+            ) && doesUserPassRollout
```
jonathannorris, line 157: swap the order of the statements here as a performance improvement.

lib/shared/bucketing/src/bucketing.ts
```diff
@@ -155,6 +155,26 @@ type SegmentedFeatureData = {
     target: PublicTarget<string>
 }
 
+const checkRollout = ({
```
jonathannorris, line 158: This function could use a more explicit name.

jonathannorris (approved):

PR #843, opened 2024-05-09
feat: add sdkConfig to track SDK Config responses

jonathannorris:
Instead of adding config metadata to `variableEvaluated` / `variableDefaulted` events like we did on the GO SDK, I think the server SDKs should generate their own `sdkConfig` event similar to the `userConfig` event we generate from the CF Worker APIs.

### DIFF
--- a/lib/shared/bucketing-assembly-script/assembly/types/dvcUser.ts
+++ b/lib/shared/bucketing-assembly-script/assembly/types/dvcUser.ts
@@ -204,6 +204,7 @@ export class DVCPopulatedUser extends JSON.Value implements DVCUserInterface {
     readonly sdkType: string
     readonly sdkVersion: string
     readonly hostname: string | null
+    readonly clientUUID: string | null
 
     constructor(user: DVCUser) {
         super()
@@ -246,6 +247,8 @@ export class DVCPopulatedUser extends JSON.Value implements DVCUserInterface {
         this.sdkType = platformData.sdkType
         this.sdkVersion = platformData.sdkVersion
         this.hostname = platformData.hostname
+        this.clientUUID = platformData.clientUUID
+
         return this
     }
 
@@ -299,6 +302,8 @@ export class DVCPopulatedUser extends JSON.Value implements DVCUserInterface {
         json.set('sdkType', this.sdkType)
         json.set('sdkVersion', this.sdkVersion)
         if (this.hostname) json.set('hostname', this.hostname)
+        if (this.clientUUID) json.set('clientUUID', this.clientUUID)
+
         return json.stringify()
     }
 }
--- a/lib/shared/bucketing-assembly-script/assembly/types/dvcUser.ts
+++ b/lib/shared/bucketing-assembly-script/assembly/types/dvcUser.ts
@@ -204,7 +204,6 @@ export class DVCPopulatedUser extends JSON.Value implements DVCUserInterface {
     readonly sdkType: string
     readonly sdkVersion: string
     readonly hostname: string | null
-    readonly clientUUID: string | null
 
     constructor(user: DVCUser) {
         super()
@@ -247,7 +246,6 @@ export class DVCPopulatedUser extends JSON.Value implements DVCUserInterface {
         this.sdkType = platformData.sdkType
         this.sdkVersion = platformData.sdkVersion
         this.hostname = platformData.hostname
-        this.clientUUID = platformData.clientUUID
 
         return this
     }
@@ -302,7 +300,6 @@ export class DVCPopulatedUser extends JSON.Value implements DVCUserInterface {
         json.set('sdkType', this.sdkType)
         json.set('sdkVersion', this.sdkVersion)
         if (this.hostname) json.set('hostname', this.hostname)
-        if (this.clientUUID) json.set('clientUUID', this.clientUUID)
 
         return json.stringify()
     }
--- a/lib/shared/config-manager/src/index.ts
+++ b/lib/shared/config-manager/src/index.ts
@@ -1,7 +1,6 @@
 import { DVCLogger } from '@devcycle/types'
 import { getEnvironmentConfig } from './request'
 import { ResponseError, UserError } from '@devcycle/server-request'
-import { DevCycleEvent, DVCPopulatedUser } from '@devcycle/js-cloud-server-sdk'
 
 type ConfigPollingOptions = {
     configPollingIntervalMS?: number
@@ -18,6 +17,8 @@ type TrackSDKConfigEventInterface = (
     url: string,
     responseTimeMS: number,
     res: Response,
+    reqEtag?: string,
+    reqLastModified?: string,
 ) => void
 
 export class EnvironmentConfigManager {
@@ -112,6 +113,8 @@ export class EnvironmentConfigManager {
         let responseError: ResponseError | null = null
         const startTime = Date.now()
         let responseTimeMS = 0
+        const reqEtag = this.configEtag
+        const reqLastModified = this.configLastModified
 
         const logError = (error: any) => {
             const errMsg =
@@ -132,8 +135,8 @@ export class EnvironmentConfigManager {
             res = await getEnvironmentConfig(
                 url,
                 this.requestTimeoutMS,
-                this.configEtag,
-                this.configLastModified,
+                reqEtag,
+                reqLastModified,
             )
             responseTimeMS = Date.now() - startTime
             projectConfig = await res.text()
@@ -175,7 +178,13 @@ export class EnvironmentConfigManager {
         }
 
         if (res && res?.status !== 304) {
-            this.trackSDKConfigEvent(url, responseTimeMS, res)
+            this.trackSDKConfigEvent(
+                url,
+                responseTimeMS,
+                res,
+                reqEtag,
+                reqLastModified,
+            )
         }
 
         if (this.hasConfig) {
--- a/lib/shared/config-manager/src/index.ts
+++ b/lib/shared/config-manager/src/index.ts
@@ -185,12 +185,12 @@ export class EnvironmentConfigManager {
                 this.configLastModified =
                     res?.headers.get('last-modified') || ''
 
-                trackEvent()
                 return
             } catch (e) {
-                trackEvent()
                 logError(new Error('Invalid config JSON.'))
                 res = null
+            } finally {
+                trackEvent()
             }
         }
 
--- a/lib/shared/config-manager/src/index.ts
+++ b/lib/shared/config-manager/src/index.ts
@@ -23,8 +23,6 @@ type TrackSDKConfigEventInterface = (
 ) => void
 
 export class EnvironmentConfigManager {
-    private readonly logger: DVCLogger
-    private readonly sdkKey: string
     private hasConfig = false
     configEtag?: string
     configLastModified?: string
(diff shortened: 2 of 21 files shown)


jonathannorris:
* **#843** <a href="https://app.graphite.dev/github/pr/DevCycleHQ/js-sdks/843?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

sdk/nodejs/src/client.ts
```diff
@@ -296,6 +303,27 @@ export class DevCycleClient {
         this.eventQueue.queueEvent(populatedUser, event)
     }
 
+    private trackSDKConfigEvent(
+        url: string,
+        responseTimeMS: number,
+        res: Response,
+    ): void {
+        this.eventQueue.queueEvent(
+            DVCPopulatedUserFromDevCycleUser({ user_id: this.clientUUID }),
```
jonathannorris, line 312: Don't have a `user_id` when the config is fetched, so just using the clientUUID as the user_id...

sdk/nodejs/src/client.ts
```diff
@@ -296,6 +303,27 @@ export class DevCycleClient {
...
+            {
+                type: 'sdkConfig',
+                target: url,
+                value: responseTimeMS,
+                metaData: {
```
JamieSinn, line 317: Do we want to include request header data too?

JamieSinn: Specifically etag/LM

jonathannorris: that's in the line below...

JamieSinn: Those are the response headers I thought, I'm saying send them both

jonathannorris: ahh good idea

JamieSinn (approved):
Test harness will probably be sad receiving more events than it expects

suthar26 (approved):

sdk/nodejs/src/client.ts
jonathannorris: @ajwootto are you good with the event format here?

lib/shared/config-manager/src/index.ts
```diff
@@ -113,6 +128,19 @@ export class EnvironmentConfigManager {
...
         }
 
+        const trackEvent = (err?: ResponseError) => {
+            if ((res && res?.status !== 304) || err) {
+                this.trackSDKConfigEvent(
+                    url,
+                    responseTimeMS,
+                    res || undefined,
+                    err,
+                    reqEtag,
```
ajwootto, line 138: shouldn't we be tracking the received etag and last modified in this event?

sdk/nodejs/src/client.ts
```diff
@@ -69,6 +71,8 @@ export class DevCycleClient {
     }
 
     constructor(sdkKey: string, options?: DevCycleServerSDKOptions) {
+        // generate UUID for this client
+        this.clientUUID = uuidv4()
```
ajwootto, line 75: do we really need to add a dependency just to generate a unique id?
we could just do something like this?
https://stackoverflow.com/a/40191779

sdk/nodejs/src/client.ts
ajwootto: I think we use the hostname as the user_id on Go. We need to decide and be consistent about it in these events

sdk/nodejs/src/client.ts
```diff
@@ -296,6 +303,35 @@ export class DevCycleClient {
...
+            DVCPopulatedUserFromDevCycleUser({ user_id: this.clientUUID }),
+            {
+                type: 'sdkConfig',
+                target: url,
+                value: responseTimeMS,
+                metaData: {
+                    clientUUID: this.clientUUID,
+                    reqEtag,
+                    reqLastModified,
+                    resEtag: res?.headers.get('etag') || undefined,
```
ajwootto, line 324: what's the purpose of `|| undefined`?

sdk/nodejs/src/client.ts
```diff
@@ -296,6 +303,35 @@ export class DevCycleClient {
...
+                    resLastModified:
+                        res?.headers.get('last-modified') || undefined,
+                    resRayId: res?.headers.get('cf-ray') || undefined,
+                    resStatus: err ? err.status : res?.status || undefined,
+                    errMsg: err ? err.message : undefined,
```
ajwootto, line 329: is this gonna be the right message? like I think this is just getting the exception message from the error object and not actually the response error data?

lib/shared/config-manager/src/index.ts
jonathannorris: yea that's taken from the `res` object.

sdk/nodejs/src/client.ts
jonathannorris: yea makes sense.

sdk/nodejs/src/client.ts
jonathannorris: @JamieSinn is that the case for the new `sdkConfig` event? I think we learned from the incident that the hostname isn't the most reliable, we could potentially combine them together.

sdk/nodejs/src/client.ts
jonathannorris: `.get()` returns `null` if it doesn't find anything, would prefer it doesn't send the param in the JSON using `undefined`.

sdk/nodejs/src/client.ts
JamieSinn: we definitely use hostname as the aggregation parameter. We can refactor it to use the UUID - but I definitely use the UUID for the sdkConfig events.

sdk/nodejs/src/client.ts
ajwootto: use `??` then

sdk/nodejs/src/client.ts
jonathannorris: should we do `${hostname}_${uuid}`? or just uuid?

JamieSinn: I would do `hostname_uuid` so we can give the customer something that correlates to their internals.

sdk/nodejs/src/client.ts
jonathannorris: yea all the response errors are wrapped in this: 
image

so `err.message` should be correct.

ajwootto: yeah but the message there is still the exception message and wouldnt contain any data about the response itself

sdk/nodejs/src/client.ts
```diff
@@ -33,6 +33,7 @@ import {
     DevCycleEvent,
 } from '@devcycle/js-cloud-server-sdk'
 import { DVCPopulatedUserFromDevCycleUser } from './models/populatedUserHelpers'
+import { randomUUID } from 'crypto'
```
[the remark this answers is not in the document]
jonathannorris: we are min 16

sdk/nodejs/src/client.ts
jonathannorris: changing the hostname that's passed into the bucketing logic to also be `hostname_uuid` to change the ones used for agg events, let me know if you think that's an issue.

JamieSinn: for agg events I think that's fine

JamieSinn: Actually - it might be easier to say `uuid@hostname` - to show that it's a given instance on a hostname? thoughts?

sdk/nodejs/src/client.ts
jonathannorris: looks like it would be the `message` field returned in the json body of the response if that's parseable: 
image

ajwootto: ok. keep in mind cloudflare doesnt have JSON responses on the CDN though, its gonna be html. So its likely this will just always say "Something went wrong"

jonathannorris: yea I think that's ok for now, I don't think we are relying on this for error tracking that way. The status is all we really need.

sdk/nodejs/src/client.ts
jonathannorris: yea that's a better format, also going to need to update the test-harness again...

JamieSinn: gotta keep us honest :P

sdk/nodejs/src/client.ts
```diff
@@ -69,6 +72,9 @@ export class DevCycleClient {
     }
 
     constructor(sdkKey: string, options?: DevCycleServerSDKOptions) {
+        // generate UUID for this client
+        this.clientUUID = randomUUID()
+        this.hostname = `${this.clientUUID}@${os.hostname()}`
```
ajwootto, line 77: shouldnt call this hostname if its not the hostname. Maybe "user_id" or "event_identifier" or something?

sdk/nodejs/src/client.ts
```diff
@@ -115,7 +124,7 @@ export class DevCycleClient {
                     platformVersion: process.version,
                     sdkType: 'server',
                     sdkVersion: packageJson.version,
-                    hostname: os.hostname(),
+                    hostname: this.hostname,
```
ajwootto, line 127: this is targeting data, you can't change this to not be the hostname

jonathannorris: I don't think it is targeting data...

jonathannorris: Yea I don't see it being used as part of targeting data, only for reporting on events.

jonathannorris: I'm happy to have you double-check, but I only see this data used for the `user_id` of `agg` events and reported as part of the populated user object for all events.

ajwootto: it's part of the automatically set user data though, I don't see why we should change the definition of hostname to not be the hostname

jonathannorris: As mentioned in the thread above, we want to try to align the user_id used by the `agg` events and this new `sdkConfig` event. I was trying to avoid making more changes to the bucketing lib, but I can separate the `uuid` from the `hostname` if needed. But it really didn't seem necessary, as the `hostname` is only ever used for reporting and not as part of any targeting rule.

ajwootto: for now maybe, it seems like an unnecessary risk to me to change the meaning of that field, and it's kinda confusing to keep track of.

lib/shared/bucketing-assembly-script/assembly/types/dvcUser.ts
```diff
@@ -299,6 +302,8 @@ export class DVCPopulatedUser extends JSON.Value implements DVCUserInterface {
         json.set('sdkType', this.sdkType)
         json.set('sdkVersion', this.sdkVersion)
         if (this.hostname) json.set('hostname', this.hostname)
+        if (this.clientUUID) json.set('clientUUID', this.clientUUID)
```
JamieSinn, line 305: why are we tying this into the populated user? It's basically unusable as a bucketing value?

jonathannorris: populated user is more for reporting rather than bucketing. My thought is that it would be useful to align the events to our `sdkConfig` event using this rather than `hostname`

ajwootto: it is most definitely for bucketing. The fields it automatically sets are used for platform and platform version filters etc

sdk/nodejs/src/client.ts
```diff
@@ -296,6 +305,35 @@ export class DevCycleClient {
...
+        reqLastModified?: string,
+    ): void {
+        const populatedUser = DVCPopulatedUserFromDevCycleUser({
+            user_id: `${this.clientUUID}@${this.hostname}`,
+        })
+
+        this.eventQueue.queueEvent(populatedUser, {
+            type: 'sdkConfig',
+            target: url,
+            value: responseTimeMS,
```
JamieSinn, line 323: given this may not be available in other sdks - do we want it as optional, or required here?

JamieSinn (approved):

sdk/nodejs/src/client.ts
jonathannorris: this is in the NodeJS SDK, it would always be available here.

lib/shared/config-manager/src/index.ts
```diff
@@ -27,17 +34,20 @@ export class EnvironmentConfigManager {
...
+    private readonly trackSDKConfigEvent: TrackSDKConfigEventInterface
 
     constructor(
         logger: DVCLogger,
         sdkKey: string,
-        setConfigBuffer: SetConfigBuffer,
+        setConfigBuffer: SetConfigBufferInterface,
         setInterval: SetIntervalInterface,
         clearInterval: ClearIntervalInterface,
+        trackSDKConfigEvent: TrackSDKConfigEventInterface,
```
ajwootto, line 50: you can combine all three of the lines declaring this field and setting it by just adding `private readonly` to the constructor

ajwootto (approved):

PR #869, opened 2024-06-05
fix: trigger event if sse url contains dvc_user

elliotCamblor:
ConfigUpdated events werent being triggered if there was a previous config, and a new config came in with a matching etag.  This is an issue for users who are subscribed to an sse channel containing "dvc_user", as the presence of that channel indicates that the user is listening for config changes triggered by self-targeting. Since self-targeting pulls in overrides separately from the CDN and then monkey-patches the config, the etag for the newly updated config will be unchanged.

## Summary by Sourcery

This pull request fixes a bug where ConfigUpdated events were not triggered for debug users if the new config had a matching etag with the previous config. It also adds tests to verify the correct behavior for both debug and non-debug users.

* **Bug Fixes**:
    - Fixed issue where ConfigUpdated events were not triggered if the new config had a matching etag with the previous config, specifically for users subscribed to an SSE channel containing 'dvc_user'.
* **Tests**:
    - Added test to ensure ConfigUpdated event is emitted for debug users even if the etag matches.
    - Added test to ensure ConfigUpdated event is not emitted for non-debug users if the etag matches.

### DIFF
--- a/sdk/js/__tests__/Client.spec.js
+++ b/sdk/js/__tests__/Client.spec.js
@@ -476,6 +476,49 @@ describe('DevCycleClient tests', () => {
             expect(variable.value).toEqual(cachedConfig.variables.key.value)
             expect(onUpdate).toBeCalledTimes(1)
         })
+
+        it('should emit configUpdated event if debug user and etag matches', async () => {
+            const debugUserTestConfig = Object.assign({}, testConfig)
+            debugUserTestConfig.etag = 'test_etag'
+            debugUserTestConfig.sse = {
+                url: 'dvc_user',
+            }
+            client = createClientWithConfigImplementation(() => {
+                return Promise.resolve(debugUserTestConfig)
+            })
+
+            const onUpdate = jest.fn()
+            client.subscribe('configUpdated', onUpdate)
+            await client.onClientInitialized()
+            // trigger 2nd config call to check if etag matches
+            await client.identifyUser({
+                user_id: 'debugUser',
+            })
+            // called once for initial config fetch and again for etag match
+            // after identifying
+            expect(onUpdate).toBeCalledTimes(2)
+        })
+
+        it('should NOT emit configUpdated event if non-debug user and etag matches', async () => {
+            const normalUserTestConfig = Object.assign({}, testConfig)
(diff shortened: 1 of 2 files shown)


sdk/js/__tests__/Client.spec.js
```diff
@@ -476,6 +476,46 @@ describe('DevCycleClient tests', () => {
...
+            })
+
+            const onUpdate = jest.fn()
+            client.subscribe('configUpdated', onUpdate)
+            await client.onClientInitialized()
+            // trigger 2nd config call to check if etag matches
+            await client.identifyUser({
+                user_id: 'debugUser',
+            })
+            expect(onUpdate).toBeCalledTimes(2)
```
suthar26, line 497: could it make more sense if the subscribe is after `onClientInitialized` and so the test would show 1 update happened vs none instead of 2 vs 1

sdk/js/src/Client.ts
```diff
@@ -700,8 +700,11 @@ export class DevCycleClient<
             config.variables,
             this.variableDefaultMap,
         )
-
-        if (!oldConfig || oldConfig.etag !== this.config.etag) {
+        // The URL including dvc_user means that this user is subscribed to a user specific ably channel
+        // This means that the user is a debug user and we should emit a config update event even if the etag
+        // is the same.
+        const isDebugUser = this.config?.sse?.url?.includes('dvc_user')
```
kaushalkapasi, line 706: looks like the config has to be defined here? removes one optional check
```suggestion
        const isDebugUser = this.config.sse?.url?.includes('dvc_user')
```

elliotCamblor: yup, this was peak laziness 😆

sdk/js/__tests__/Client.spec.js
elliotCamblor: Probably doesnt hurt to check that its getting called when there's no `oldConfig` yet as well

suthar26: just got confused for a second, maybe a comment that explains that? could just be me 😅

suthar26 (approved):
code looks good

kaushalkapasi (approved):

PR #874, opened 2024-06-10
fix: next pages strict mode and options

ajwootto:
- remove the "synchronization" logic from the pages router SDK because the SSR data isn't "reactive", it just gets sent once on the client
- just pass it in to the initialization of the DevCycle SDK

### DIFF
--- a/sdk/nextjs/src/pages/appWithDevCycle.tsx
+++ b/sdk/nextjs/src/pages/appWithDevCycle.tsx
@@ -1,8 +1,8 @@
 import type { AppProps as NextJsAppProps } from 'next/app'
 import hoistNonReactStatics from 'hoist-non-react-statics'
 import { SSRProps } from './types'
-import { DevCycleProvider, useDevCycleClient } from '@devcycle/react-client-sdk'
-import React, { useEffect, useRef } from 'react'
+import { DevCycleProvider } from '@devcycle/react-client-sdk'
+import React from 'react'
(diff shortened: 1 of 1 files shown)


sdk/nextjs/src/pages/appWithDevCycle.tsx
```diff
@@ -87,21 +39,25 @@ export const appWithDevCycle = <Props extends NextJsAppProps>(
             <DevCycleProvider
                 config={{
                     sdkKey: devcycleSSR.sdkKey,
+                    user: devcycleSSR.user,
                     options: {
                         ...additionalOptions,
+                        ...(onServerside
```
jsalaber, line 45: can this be on top of additional options? if the user defines the `disableAutomaticEventLogging` option it'll get overwritten here

ajwootto: it has to be overwritten on the serverside, because event logging there isn't supported

jsalaber: ah okay I see it won't override it if the onServerSide is false on the client now

jsalaber (approved):

sdk/nextjs/src/pages/appWithDevCycle.tsx
ajwootto: yup

PR #890, opened 2024-07-04
chore: add resolve alias to webpack to fix example app dynamic imports

ajwootto:
Fixes nodejs example app


elliotCamblor (approved):
wtf is this really all it takes?

ajwootto:
> wtf is this really all it takes?

yeah, the problem is that in the living hell that is ESM Typescript, you have to specify `.js` extensions on import statements for them to work in ESM, even if the "original" file is Typescript. The Typescript compiler knows how to deal with that, but Webpack doesn't (at least the version we're using), so we basically have to tell it "if you see an import for a `.js` file, also consider importing `.ts` files of the same name while bundling".

PR #891, opened 2024-07-04
fix: Update streaming connection URL when config received

emmawillis:


suthar26 (approved):
lgtm

sdk/js/src/Client.ts
```diff
@@ -707,6 +693,30 @@ export class DevCycleClient<
...
+                        this.onSSEMessage.bind(this),
+                        this.logger,
+                    )
+                } else {
+                    this.logger.info(
+                        'Disabling Realtime Updates based on Initialization parameter',
+                    )
+                }
+            }
+            if (
```
ajwootto, line 713: should this be an else if? otherwise it might open a new streaming connection and then immediately reconnect it

sdk/js/src/StreamingConnection.ts
```diff
@@ -11,6 +11,12 @@ export class StreamingConnection {
         this.openConnection()
     }
 
+    public updateURL(url: string): void {
+        this.url = url
```
ajwootto, line 15: might be better to change this property after you call close()

sdk/js/src/Client.ts
emmawillis: true good point, fixed

PR #896, opened 2024-07-17
feat: Vercel Edge Config Adapter

ajwootto:
- create package for retrieving configs from Vercel Edge Config on Node.js
- refactor config manager to accept a generic "storage provider" so the source of the config can be swapped


project.json
```diff
@@ -0,0 +1,14 @@
+{
+    "name": "devcycle-js-sdks",
```
jonathannorris, line 2: what is file this for?

ajwootto: its a file that nx keeps automatically adding every time you generate a new package. I was thinking of just giving up and letting it be generated...

jonathannorris (approved):

PR #897, opened 2024-07-18
feat: add edge config support to nextjs

ajwootto:


lib/shared/config-manager/src/ConfigSource.ts
```diff
@@ -10,20 +10,31 @@ export abstract class ConfigSource {
...
      * @param sdkKey
      * @param kind
+     * @param obfuscated
      * @param lastModifiedThreshold
      */
     abstract getConfig(
         sdkKey: string,
         kind: 'server' | 'bootstrap',
+        obfuscated: boolean,
         lastModifiedThreshold?: string,
```
jonathannorris, line 20: I wonder if we should just move all these params into a shared type rather than keep passing them all around?

jonathannorris (approved):

PR #903, opened 2024-07-23
wrap vercel flags

ajwootto:


e2e/nextjs/app-router/app/app/normal/devcycle.ts
```diff
@@ -20,3 +22,8 @@ export const {
     },
     options: { enableStreaming: false },
 })
+
+export const getFlag = setupDevCycleVercelFlagHelper({
```
jonathannorris, line 26: Why as a separate import and not a part of `setupDevCycle()`?

ajwootto: because it depends on the `@vercel/flags` package which I wanted to make an optional dependency. So the idea is that if you want to use it with Vercel, you install their flags package and import this helper from `@devcycle/nextjs-sdk/vercel` where it expects the package to be available.

sdk/nextjs/vercel.ts
```diff
@@ -0,0 +1,98 @@
+import { unstable_flag as flag } from '@vercel/flags/next'
+import { setupDevCycle } from './src/server/setupDevCycle'
+
+// mimic the Vercel JsonValue type since it is not exported
+type JsonValue =
```
jonathannorris, line 5: does the array support here get pulled out somewhere else?

ajwootto: what do you mean?

sdk/nextjs/vercel.ts
```diff
@@ -0,0 +1,98 @@
...
+        })
+
+        if (!featureForVariable) {
+            console.error('[DevCycle] No Feature found for Variable', key)
+            return defaultValue
+        }
+
+        const getFlag = flag<string | null>({
+            key: featureForVariable.key,
+            defaultValue: null,
```
jonathannorris, line 60: why no default here?

ajwootto: because this is serving the variation of the feature that should be used to obtain the flag value. The default for that should be "no variation" which is interpreted by the code further down as "use the variable's default"

jonathannorris (approved):

jonathannorris:
@ajwootto is this PR still valid?

ajwootto:
> @ajwootto is this PR still valid?

no

Issue #913, opened 2024-08-06
`initializeDevCycle.onClientInitialized()` does not throw error when connection to DVC APIs fail


doblinger-extron:
I have recently included DevCycle in my application and experienced unexpected behavior.
My expectation is that the client should not initialize if DVC cannot be reached or an invalid client SDK key is provided.

#### Scenario 1
An invalid SDK key is provided to the client. 

#### Scenario 2
The DVC server cannot be reached. I was able to reproduce this by blocking network requests to DVC servers in Chrome devtools.

#### Example code
```typescript
try {
  const client = await initializeDevCycle('invalid-client-id', {}).onClientInitialized();
  console.log('Is client initialized: ' + client.isInitialized);
} catch (error) {
  console.log('An error occurred');
}
```

### Expected output:
`'An error occurred' // console.log`

### Actual output:
`'Is client initialized: true' // console.log`
Further more, because the client initializes successfully, when `await client.identifyUser(user)` is invoked, an unhandled exception occurs -- seemingly in the background? -- and my application hangs.

I've been able to temporarily workaround the issue by pairing disabling of the config cache as well as verifying the `client.config.project` object is undefined, but this is certainly hacky.

That code looks something like 
```typescript
const client = await initializeDevCycle('invalid-client-id', { disableConfigCache: true }).onClientInitialized();

// only identify the user when project is truthy
if (client.config?.project) {
  await client.identifyUser(user);
}
```

kaushalkapasi:
Hey @doblinger-extron, thanks for raising this. We'll investigate the issues raised and get back to you.

doblinger-extron:
@kaushalkapasi Thanks for investigating!

As far as invalid SDK keys goes, I tried multiple combinations, such as a random string like `invalid-sdk-key` as well as just adding an extra character, i.e. `1`, to the end of a valid key.

I forgot to mention that I have version `1.28.1` of `@devcycle/js-client-sdk` installed. I haven't tried other versions of the sdk.

kaushalkapasi:
@doblinger-extron quick update, I've verified the issue you described in **Scenario 1** (invalid SDK Key) and that is definitely a bug / not the intended behaviour. We'll be making an update to fix that.

[...]

PR #918, opened 2024-08-07
fix: js-sdk was improperly setting isIntialized to true even for situations where the SDK token was invalid

kaushalkapasi:
Reference: https://github.com/DevCycleHQ/js-sdks/issues/913

- fix: adds initialization validation to ensure the SDK key provide to the JS SDK starts with `dvc_client` or `client`
  - fix: allows server key to be used when initializing the JS Client SDK from the NextJS SDK
- fix: calls the reject handler for on Initialized if the Client SDK API rejects the SDK token with a `401 - Forbidden` error message


sdk/js/src/Client.ts
```diff
@@ -207,6 +206,7 @@ export class DevCycleClient<
                     Date.now(),
                 )
             }
+            this._isInitialized = true
```
ajwootto, line 209: this seems to be doing something different than the title of the PR. Are we trying to make this false when the SDK key is invalid, or when the config request fails for any reason?

sdk/js/src/Client.ts
```diff
@@ -122,7 +122,6 @@ export class DevCycleClient<
 
         this.onInitialized = new Promise((resolve, reject) => {
             this.resolveOnInitialized = (value) => {
-                this._isInitialized = true
```
ajwootto, line 125: this promise should reject if the key is invalid

sdk/js/src/Client.ts
kaushalkapasi: ah, that's right.. this should only be when the config-key is invalid.. will update

sdk/js/src/index.ts
```diff
@@ -97,6 +97,12 @@ export function initializeDevCycle<
         throw new Error('Missing SDK key! Call initialize with a valid SDK key')
     }
 
+    if (!sdkKey.startsWith('client') && !sdkKey.startsWith('dvc_client')) {
+        throw new Error(
+            'Invalid SDK key! SDK key must start with "client_" or "dvc_client_"',
```
ajwootto, line 102: just use the error message format from the node sdk

sdk/js/src/Request.ts
```diff
@@ -66,12 +66,16 @@ export const getConfigJson = async (
...
         return await res.json()
     } catch (e) {
+        const error = e as ResponseError
         logger.error(
             `Request to get config failed for url: ${url}, ` +
-                `response message: ${e}`,
+                `response message: ${error.message}`,
         )
+        if (error.status === 401 || error.status === 403) {
+            throw new Error(`Invalid SDK Key. Error details: ${error.message}`)
```
ajwootto, line 75: you should use the `UserError` pattern we have in the Node SDK here rather than inspecting the error message contents.

sdk/js/src/Client.ts
```diff
@@ -121,9 +124,14 @@ export class DevCycleClient<
         this.registerVisibilityChangeHandler()
 
         this.onInitialized = new Promise((resolve, reject) => {
-            this.resolveOnInitialized = (value) => {
-                this._isInitialized = true
-                resolve(value)
+            this.resolveOnInitialized = (value, error) => {
```
ajwootto, line 127: this function should probably be called `settleOnInitialized` to better reflect what it does now

sdk/js/src/Request.ts
kaushalkapasi: done! also updated the invalid sdk key errors to be `UserError`

sdk/js/__tests__/Client.spec.js
```diff
@@ -316,16 +326,28 @@ describe('DevCycleClient tests', () => {
...
+        })
+
+        it('should not initialize if there is an error reaching the DevCycle Client SDK API', async () => {
+            const client = new DevCycleClient(error_test_key, {
+                user_id: 'user1',
+            })
+            const callback = jest.fn()
+            const onInitialized = client.onClientInitialized(callback)
+            expect(onInitialized).toBeFalsy()
+            expect(client.isInitialized).toBe(false)
```
ajwootto, line 350: i thought we wanted this to be true?

sdk/js/src/utils.ts
```diff
@@ -86,6 +86,15 @@ export function checkIsServiceWorker(): boolean {
         self.constructor.name === 'ServiceWorkerGlobalScope'
     )
 }
+
+export class UserError extends Error {
+    constructor(error: Error | string) {
+        super(error instanceof Error ? error.message : error)
+        this.name = 'UserError'
+        this.stack = error instanceof Error ? error.stack : undefined
```
ajwootto, line 94: wont this wipe the original stack for the error?

sdk/js/__tests__/Client.spec.js
kaushalkapasi: ah the test description is wrong.. but this test is with a bad SDK Key, thus the `isInitialized` should be false

ajwootto: thats not what its testing though, since youre just throwing a regular error

sdk/js/src/utils.ts
kaushalkapasi: I don't believe so, this is the same implementation as the Node SDK

ajwootto: well it looks like its reassigning the stack property to be undefined

sdk/js/src/Request.ts
```diff
@@ -66,12 +70,18 @@ export const getConfigJson = async (
         const res = await getWithTimeout(url, requestConfig, 5000)
         return await res.json()
     } catch (e) {
+        const error = e as ResponseError
```
ajwootto, line 73: this isnt a valid type assertion, error could be anything (like what about when json fails to parse?)

kaushalkapasi: will refactor 👍

ajwootto (approved):

PR #919, opened 2024-08-08
feat: add comparators to ConfigBody

suthar26:
- add comparator types to the shared types


lib/shared/types/src/types/config/models/audience.ts
```diff
@@ -12,6 +12,10 @@ export enum FilterComparator {
     '!exist' = '!exist',
     'contain' = 'contain',
     '!contain' = '!contain',
+    'startsWith' = 'startsWith',
```
elliotCamblor, line 15: its `startWith` and `endWith`, not `startsWith` `endsWith`

elliotCamblor (approved):

PR #921, opened 2024-08-09
chore: fixed passing in logger object when stringifying to prevent circular json error

jsalaber:
# Changes

- filtered out the `logger` option when initializing the event queue in the bucketing library to avoid JSON.stringifying a logger that could be passed in, e.g. `winston`. If the logger has a circular dependency, stringifying it will cause an error. Since the event queue options don't utilize the logger anyway, only pass in the options it expects

### DIFF
--- a/sdk/nodejs/src/eventQueue.ts
+++ b/sdk/nodejs/src/eventQueue.ts
@@ -106,10 +106,16 @@ export class EventQueue {
             this.eventFlushIntervalMS,
         )
 
+        const eventQueueOptions = {
+            eventRequestChunkSize: chunkSize,
+            disableAutomaticEventLogging: options.disableAutomaticEventLogging,
+            disableCustomEventLogging: options.disableCustomEventLogging,
+        }
+
         this.bucketing.initEventQueue(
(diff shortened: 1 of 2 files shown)


suthar26 (approved):

sdk/nodejs/src/eventQueue.ts
```diff
@@ -106,10 +106,16 @@ export class EventQueue {
             this.eventFlushIntervalMS,
         )
 
+        const eventQueueOptions = {
+            eventRequestChunkSize: chunkSize,
```
ajwootto, line 110: how did this work before if you're renaming this property? or did it just not work

jsalaber: it unwrapped it from the options as a different name `chunkSize`, the original property name on the options is `eventRequestChunkSize`

jsalaber: although this does technically change it so that the chunk size is now defaulted to 100 defined above, not sure if that's what we want

ajwootto (approved):

PR #928, opened 2024-08-13
feat: add new comparators to bucketing lib

suthar26:
add `startWith` `endWith` and the opposites to the bucketing lib

### DIFF
--- a/lib/shared/bucketing/__tests__/segmentation.test.ts
+++ b/lib/shared/bucketing/__tests__/segmentation.test.ts
@@ -1091,6 +1091,132 @@ describe('SegmentationManager Unit Test', () => {
                 segmentation.checkStringsFilter('helloWorld', filter),
             )
         })
+        it('should return true if starts with filter and value starts with', () => {
+            const filter = {
+                type: 'user',
+                comparator: 'startWith',
+                values: ['testuser@'],
+            } as AudienceFilterOrOperator
+            assert.strictEqual(
+                true,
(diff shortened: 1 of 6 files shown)


lib/shared/bucketing/__tests__/segmentation.test.ts
```diff
@@ -1091,6 +1091,90 @@ describe('SegmentationManager Unit Test', () => {
...
+                values: [''],
+            } as AudienceFilterOrOperator
+            assert.strictEqual(
+                false,
+                segmentation.checkStringsFilter(
+                    'testuser@devcycle.com',
+                    filter,
+                ),
+            )
+        })
```
JamieSinn, line 1177: Can we add this for all the new filters and contains?

suthar26: do you mean add it to `!startWith` `!endWith` ?

JamieSinn: yep - and the contains operator - just so we have a source of truth with regard to what we expect

suthar26: added, do you want to make sure its expecting the correct values?

suthar26: yeah added the comparators

JamieSinn (approved):

PR #930, opened 2024-08-14
fix: disable last modified check on vercel edge config and fix type definitions

ajwootto:
- make sure edge config always returns a config in Nextjs by disabling the last modified check in that SDK
- fix the type definitions so that the edge config client is compatible with Nextjs types

### DIFF
--- a/lib/shared/vercel-edge-config/src/edge-config.ts
+++ b/lib/shared/vercel-edge-config/src/edge-config.ts
@@ -7,12 +7,14 @@ export class EdgeConfigSource extends ConfigSource {
         super()
     }
 
-    async getConfig(
+    async getConfig<T extends boolean>(
         sdkKey: string,
         kind: 'server' | 'bootstrap',
         obfuscated: boolean,
+        lastModifiedThreshold?: string,
+        skipLastModified?: T,
     ): Promise<{
-        config: ConfigBody | null
+        config: T extends true ? ConfigBody : ConfigBody | null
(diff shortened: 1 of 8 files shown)


lib/shared/vercel-edge-config/src/edge-config.ts
```diff
@@ -7,12 +7,14 @@ export class EdgeConfigSource extends ConfigSource {
         super()
     }
 
-    async getConfig(
+    async getConfig<T extends boolean>(
         sdkKey: string,
         kind: 'server' | 'bootstrap',
         obfuscated: boolean,
+        lastModifiedThreshold?: string,
```
kaushalkapasi, line 14: where's this used?

ajwootto: it's not used here but its part of the type signature for the abstract class

sdk/nextjs/src/pages/bucketing.ts
```diff
@@ -49,6 +48,8 @@ export const getBucketedConfig = async (
         sdkKey,
         'bootstrap',
         obfuscated,
+        '',
```
kaushalkapasi, line 51: so we're no longer allowing the `lastModifiedThreshold` to be passed in?

ajwootto: it never was

kaushalkapasi (approved):

PR #931, opened 2024-08-14
feat: change vercel path to config v2

ajwootto:


elliotCamblor (approved):
I'd also like to make sure we aren't uploading both v1 and v2 configs to vercel, but I can fix that before we turn on the ff

PR #937, opened 2024-08-16
feat: use bucketingKey from users custom data

jsalaber:
# Changes

- use `bucketingKey` if user has it defined in `customData`, otherwise use `"null"` to bucket users that don't have a customData value

### DIFF
--- a/lib/shared/bucketing/__tests__/bucketing.test.ts
+++ b/lib/shared/bucketing/__tests__/bucketing.test.ts
@@ -10,6 +10,7 @@ import {
     config,
     barrenConfig,
     configWithNullCustomData,
+    configWithBucketingKey,
 } from '@devcycle/bucketing-test-data'
 
 import moment from 'moment'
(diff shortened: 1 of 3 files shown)


lib/shared/bucketing/src/bucketing.ts
```diff
@@ -326,3 +331,26 @@ export const generateBucketedConfig = ({
...
+export const getUserValueForBucketingKey = ({
+    user,
+    target,
+}: {
+    user: DVCBucketingUser
+    target: PublicTarget
+}): string => {
+    if (target.bucketingKey && target.bucketingKey !== 'user_id') {
+        const bucketingValue =
+            user.customData?.[target.bucketingKey] || DEFAULT_BUCKETING_VALUE
```
kaushalkapasi, line 344: this should also check against the `user.privateCustomData`

lib/shared/bucketing/__tests__/bucketing.test.ts
```diff
@@ -781,6 +782,84 @@ describe('Config Parsing and Generating', () => {
...
+        const bucketedConfigSameUserDifferentFood = generateBucketedConfig({
+            config: configBucketedByFood,
+            user: sameUserDifferentFood,
+        })
+        const bucketedConfigDifferentUserSameFood = generateBucketedConfig({
+            config: configBucketedByFood,
+            user: differentUserSameFood,
+        })
+        expect(bucketedConfigOrig).not.toEqual(bucketedConfigByFood)
+        expect(bucketedConfigByFood).not.toEqual(
```
elliotCamblor, line 855: might be worth double checking that these `not toEqual` assertions wont flake. 
It should be fine because I think the data is completely static, but in theory having different bucketing mechanisms doesn't guarantee that the configs will be different

lib/shared/bucketing/src/bucketing.ts
elliotCamblor: seconding this, the `privateCustomData` might already have been merged with `customData` at this point though

lib/shared/bucketing/__tests__/bucketing.test.ts
jsalaber: made them check a feature I added in the test data that has a bunch of different variations and check that they don't match (had to test a couple of times that they don't match, they should be consistently pass moving forward)

elliotCamblor: coolio

elliotCamblor (approved):

PR #939, opened 2024-08-19
fix: reset current promise after every config request

kaushalkapasi:
- fix: reset `currentPromise` in the config request consolidation queue, including if there is an error while fetching

### DIFF
--- a/sdk/js/src/ConfigRequestConsolidator.ts
+++ b/sdk/js/src/ConfigRequestConsolidator.ts
@@ -75,6 +75,9 @@ export class ConfigRequestConsolidator {
             .catch((err) => {
                 resolvers.forEach(({ reject }) => reject(err))
             })
+            .finally(() => {
+                this.currentPromise = null
+            })
 
         if (this.resolvers.length) {
             this.processQueue()
@@ -90,7 +93,6 @@ export class ConfigRequestConsolidator {
         )
         this.requestParams = null
         const bucketedConfig = await this.currentPromise
-        this.currentPromise = null
         return bucketedConfig
     }
 }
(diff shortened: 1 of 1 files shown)


sdk/js/src/ConfigRequestConsolidator.ts
```diff
@@ -75,6 +75,9 @@ export class ConfigRequestConsolidator {
             .catch((err) => {
                 resolvers.forEach(({ reject }) => reject(err))
             })
+            .finally(() => {
+                this.currentPromise = null
```
ajwootto, line 79: any concerns about what changing the order of these operations will do?

kaushalkapasi: since the `currentPromise` will not execute if there's already an in-flight request, I don't see any changes with this format... the other option is to explicitly move this statement within the `catch` (line 75) and leave the original code as-is, which would eliminate the `.finally` here

ajwootto: well now you're resolving all the waiting resolvers before setting currentPromise back to null, whereas before it was null first

kaushalkapasi: hm, i'll switch it to set it to null before the resolvers.. because if there's any errors in the resolvers, then it could potentially delay subsequent requests

kaushalkapasi: updated to explicitly reset the `currentPromise` in the catch handler

ajwootto: oh nvm diff is weird

ajwootto: it would be cleaner to just try catch inside performRequest

kaushalkapasi: sounds good, that way we can avoid having the reset behaviour split

ajwootto (approved):

PR #940, opened 2024-08-19
feat: added bucketing key logic to shared-bucketing-as

jsalaber:
# Changes

- added tests for bucketing key 
- used bucketing key if it exists instead of user_id

### DIFF
--- a/lib/shared/bucketing-assembly-script/assembly/bucketing/bucketing.ts
+++ b/lib/shared/bucketing-assembly-script/assembly/bucketing/bucketing.ts
@@ -1,31 +1,32 @@
 import { JSON } from '@devcycle/assemblyscript-json/assembly'
 import { first, last } from '../helpers/lodashHelpers'
 import {
-    ConfigBody,
-    Target as PublicTarget,
-    Feature as PublicFeature,
+    ConfigBodyV2 as ConfigBody,
+    TargetV2 as PublicTarget,
+    FeatureV2 as PublicFeature,
     BucketedUserConfig,
     Rollout as PublicRollout,
     DVCPopulatedUser,
     SDKVariable,
     SDKFeature,
     RolloutStage,
-    Target,
+    TargetV2 as Target,
     Variation,
     FeatureVariation,
-    Feature,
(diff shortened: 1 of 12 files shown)


lib/shared/bucketing-assembly-script/assembly/bucketing/bucketing.ts
```diff
@@ -381,3 +385,35 @@ export function _generateBucketedVariableForUser(
...
+    if (target.bucketingKey && target.bucketingKey !== 'user_id') {
+        let bucketingValue: string = DEFAULT_BUCKETING_VALUE
+        const customData = user.customData
+        const privateCustomData = user.privateCustomData
+        if (customData) {
+            const customDataValue = getValueFromJSONOptional(customData, target.bucketingKey)
+            bucketingValue = customDataValue
+                ? customDataValue.toString()
+                : DEFAULT_BUCKETING_VALUE
+        } else if (privateCustomData) {
```
suthar26, line 402: should this not be checked when customDataValue is missing?

elliotCamblor: The `User` class has a `getCombinedCustomData` function that should allow you to abstract away the private vs public custom data

jsalaber: much better ty

lib/shared/bucketing-assembly-script/assembly/types/target.ts
```diff
@@ -18,6 +18,7 @@ export class Target extends JSON.Value {
     readonly _audience: Audience
     readonly rollout: Rollout | null
     readonly distribution: TargetDistribution[]
+    readonly bucketingKey: string
```
elliotCamblor, line 21: does this class need to be updated, or does it make more sense to use the V2 classes I created? I did it that way so that we could validate v1 configs against the v1 classes in the config service. 

Technically this wont fail for v1 configs since they all have bucketing key anyways, so its more of a philosophical question than anything

elliotCamblor (approved):

PR #941, opened 2024-08-20
chore: use v2 configs in nodejs

jsalaber:
# Changes

- use v2 configs in nodejs


elliotCamblor:
we'll need to make sure the version of the wasm bucketing lib is the latest version before we make this change, otherwise the sdk will be pulling v2 configs without v2 functionality

jsalaber:
@elliotCamblor is there anyway to do that in runtime or will we just need to check manually

elliotCamblor (approved):

PR #957, opened 2024-09-16
chore: update protobufjs

jonathannorris:

### DIFF
--- a/.github/workflows/cf-worker-example-test.yml
+++ b/.github/workflows/cf-worker-example-test.yml
@@ -7,6 +7,7 @@ jobs:
     build:
         runs-on:
             labels: ubuntu-latest-4-core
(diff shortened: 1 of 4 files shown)


JamieSinn:
https://github.com/DevCycleHQ/js-sdks/blob/main/package.json#L62
Not updating the package.json?

jonathannorris:
> https://github.com/DevCycleHQ/js-sdks/blob/main/package.json#L62 Not updating the package.json?

Doesn't need to be as its using `^` which allow for minor changes. But yea I guess we should on these SDKs

JamieSinn:
my thought is that currently if there's a compatibility issue - we're still exposed based on version matching

JamieSinn (approved):

PR #958, opened 2024-09-17
chore: update bucketing distribution tests for numbers with 4 decimal places

kaushalkapasi:
- chore: update WASM bucketing test data to include config with 4 decimal places for a random distribution 
- chore: update JS bucketing test to include config with 4 decimal places for a random distribution

### DIFF
--- a/lib/shared/bucketing/__tests__/bucketing.test.ts
+++ b/lib/shared/bucketing/__tests__/bucketing.test.ts
@@ -25,13 +25,15 @@ describe('User Hashing and Bucketing', () => {
             _audience: { _id: 'id', filters: [] } as unknown as Audience,
             _id: 'target',
             distribution: [
-                { _variation: 'var1', percentage: 0.25 },
-                { _variation: 'var2', percentage: 0.45 },
+                { _variation: 'var1', percentage: 0.2555 },
+                { _variation: 'var2', percentage: 0.4445 },
                 { _variation: 'var4', percentage: 0.2 },
                 { _variation: 'var3', percentage: 0.1 },
             ],
         }
-        times(30000, () => {
+
+        // run 100,000 times to get a good distribution
+        times(100000, () => {
             const user_id = uuid.v4()
             const { bucketingHash } = generateBoundedHashes(
                 user_id,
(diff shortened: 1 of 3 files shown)


lib/shared/bucketing/__tests__/bucketing.test.ts
```diff
@@ -25,8 +25,8 @@ describe('User Hashing and Bucketing', () => {
             _audience: { _id: 'id', filters: [] } as unknown as Audience,
             _id: 'target',
             distribution: [
-                { _variation: 'var1', percentage: 0.25 },
-                { _variation: 'var2', percentage: 0.45 },
+                { _variation: 'var1', percentage: 0.2555 },
```
jonathannorris, line 28: you should probably update the expected value ranges below.

kaushalkapasi: updated to narrow the expected range to be 0.5% around `var1` and `var2`

kaushalkapasi: @jonathannorris thoughts on the ranges.. these tests will flake even with 50k evals due to random distributions being, well random 😅

jonathannorris: @kaushalkapasi has that flakeness changed? it shouldn't be that bad...

kaushalkapasi: nope.. with 4 decimal places and the narrowing of the range checks (was previously  +/- 1% of the distribution percentage, now its +/- 0.5%) there's just a higher likely hood of things being just outside that +/- 0.5%

kaushalkapasi: the existing tests cover that the config is saved and parsed correctly.. should we add an explicit check for the 4 decimal places? or is the existing test that the config is valid / matches the input config when parsed out of wasm enough?

jonathannorris: does that test config have these longer decimal places in them?

jonathannorris (approved):

lib/shared/bucketing/__tests__/bucketing.test.ts
kaushalkapasi: yeah, I've updated the underlying `lib/shared/bucketing-test-data/src/data/testData.ts` file to include a couple of instances to use numbers with 4 decimal places

PR #974, opened 2024-10-07
support self targeting in app router

ajwootto:


jonathannorris (approved):

sdk/nextjs/src/server/requests.ts
```diff
@@ -19,3 +22,56 @@ export const fetchCDNConfig = async (
...
+
+export const sdkConfigAPI = async (
+    sdkKey: string,
+    obfuscated: boolean,
+    user: DVCPopulatedUser,
+): Promise<Response> => {
+    return await fetch(getSDKAPIUrl(sdkKey, obfuscated, user), {
+        next: {
+            revalidate: 60,
+            tags: [sdkKey, user.user_id],
```
elliotCamblor, line 50: so does this mean that the config request will be cached for the user regardless of if their user data changes? (aside from userid of course)

ajwootto: It's only cached if the entire contents of the request are the same. So if the user data changes that will change the query params so it won't be cached

PR #978, opened 2024-10-23
fix: fixed retrying logic for invalid sdk key

jsalaber:
# Changes

- If the sdk key is invalid, close the event queue

### DIFF
--- a/sdk/nodejs/src/eventQueue.ts
+++ b/sdk/nodejs/src/eventQueue.ts
@@ -49,6 +49,7 @@ export class EventQueue {
     eventFlushIntervalMS: number
     flushEventQueueSize: number
     maxEventQueueSize: number
+    disabledEventFlush: boolean
     private flushInterval: NodeJS.Timer
     private flushInProgress = false
     private flushCallbacks: Array<(arg: unknown) => void> = []
@@ -63,6 +64,8 @@ export class EventQueue {
         this.reporter = options.reporter
         this.eventsAPIURI = options.eventsAPIURI
         this.eventFlushIntervalMS = options?.eventFlushIntervalMS || 10 * 1000
+        this.disabledEventFlush = false
+
         if (this.eventFlushIntervalMS < 500) {
             throw new Error(
                 `eventFlushIntervalMS: ${this.eventFlushIntervalMS} must be larger than 500ms`,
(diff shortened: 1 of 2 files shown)


sdk/nodejs/src/eventQueue.ts
```diff
@@ -219,6 +220,12 @@ export class EventQueue {
                     this.logger.debug(
                         `DevCycle Error Flushing Events response message: ${ex.message}`,
                     )
+                    if (ex instanceof ResponseError && ex.status === 401) {
+                        this.logger.debug(
+                            `SDK key is invalid, stopping retrying`,
+                        )
+                        return
```
ajwootto, line 227: I don't think this would do anything except put the bucketing events queue in a limbo state where the status of the current payload is never updated from pending, while new events continue to accumulate and the flush interval continues to run/

ajwootto: there also isn't any "retrying" here, this is just the flush loop that runs on a fixed interval. So the bigger mystery is why it was tight-looping request retries yesterday

sdk/nodejs/src/eventQueue.ts
```diff
@@ -121,6 +124,8 @@ export class EventQueue {
 
     cleanup(): void {
         clearInterval(this.flushInterval)
+        this.bucketing.flushEventQueue(this.sdkKey)
```
ajwootto, line 127: what's the idea behind calling this here?

jsalaber: basically to flush out anything in the bucketing queue so it doesn't stay there, it doesn't actually make a request it just doesn't use the return data in anything, it just throws it away

ajwootto: might be worth adding a comment about that

ajwootto: also I dont think this will actually have any effect on memory since this just moves the events to a different internal queue inside the bucketing lib

jsalaber: hmm yeah, I've thought about closing the event queue but that just causes problems when calling variables since it tries to get the event queue when evaluating a variable

jonathannorris:
@jsalaber what's the status of this PR?

ajwootto (approved):

PR #982, opened 2024-10-24
fix: update SDK for next15 compatibility

ajwootto:
- update next SDK to use the new async version of the headers() function for next 15
- remove unused cookies utility


sdk/nextjs/src/server/initialize.ts
```diff
@@ -72,10 +72,10 @@ export const initialize = async (
     }
 
     if (!initializeAlreadyCalled) {
-        client.synchronizeBootstrapData(config, user, getUserAgent(options))
+        client.synchronizeBootstrapData(config, user, await getUserAgent(options))
```
suthar26, line 75: is this not part of the same function as where `userAgent` is being called? seems like another await for no reason..

suthar26 (approved):

jsalaber (approved):

PR #984, opened 2024-10-24
chore: OIDC Secrets

JamieSinn:


jonathannorris (approved):

.github/workflows/cf-worker-example-test.yml
```diff
@@ -23,7 +23,7 @@ jobs:
             - name: Run Yarn
               run: yarn --immutable
             - name: Setup .dev.vars file
-              run: echo "DEVCYCLE_SERVER_SDK_KEY=${{ secrets.DEVCYCLE_SERVER_SDK_KEY }}" > dev-apps/js-cloud-server/cloudflare-worker/.dev.vars
+              run: echo "DEVCYCLE_SERVER_SDK_KEY=dvc_server_token_hash" > dev-apps/js-cloud-server/cloudflare-worker/.dev.vars
```
ajwootto, line 26: why is this being replaced with a dummy value?

JamieSinn: Because it's our public test token - where there is no need to have an actual prod token for this test

ajwootto: but it's checking specific variable results from the app that is executed

JamieSinn: No? it's checking that a header exists? Or do you mean the actual test app

JamieSinn: From the testing - it passed?

ajwootto: is it testing what was intended to be tested by this? like I don't understand why we wouldnt use real tokens in an e2e test...

JamieSinn: I think that's a larger question on the test itself outside the scope of this PR. The test is not checking anything meaningful- no; which is why the token change as part of our update is not critical

ajwootto: its not outside the scope you're the one changing it.

PR #989, opened 2024-11-07
fix: attach sync call directly to promise

ajwootto:
- handle synchronizing to Nextjs client by attaching a handler to the promise directly

### DIFF
--- a/sdk/nextjs/src/client/internal/InternalDevCycleClientsideProvider.tsx
+++ b/sdk/nextjs/src/client/internal/InternalDevCycleClientsideProvider.tsx
@@ -1,13 +1,5 @@
 'use client'
-import React, {
-    Suspense,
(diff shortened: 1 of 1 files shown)


sdk/nextjs/src/client/internal/InternalDevCycleClientsideProvider.tsx
```diff
@@ -1,13 +1,5 @@
...
-import React, {
-    Suspense,
-    use,
-    useCallback,
-    useContext,
-    useEffect,
-    useRef,
-    useState,
-} from 'react'
+import React, { Suspense, use, useRef } from 'react'
```
jsalaber, line 2: Suspense can be removed

jsalaber (approved):

PR #1001, opened 2024-11-27
feat: support strong typing of variable keys

SimeonC:
This allows better strong typing of the string values in the react hooks.
There are no changes to the js SDK but due to the usage of generics all the other libraries should get advantages from this too.

CleanShot 2024-11-27 at 16 41 57

Please do let me know if I've missed something as this was mostly tested out by playing around with the types in node_modules. I think that if this lands we could also update the cli to include the definition output as an option?

### DIFF
--- a/sdk/js-cloud-server/src/request.ts
+++ b/sdk/js-cloud-server/src/request.ts
@@ -2,6 +2,7 @@ import { DVCPopulatedUser } from './models/populatedUser'
 import { DevCycleEvent } from './types'
 import { DevCycleServerSDKOptions } from '@devcycle/types'
 import { post } from '@devcycle/server-request'
+import { VariableKey } from '@devcycle/js-client-sdk'
 
 export const HOST = '.devcycle.com'
 
@@ -58,7 +59,7 @@ export async function getAllVariables(
 export async function getVariable(
     user: DVCPopulatedUser,
     sdkKey: string,
-    variableKey: string,
+    variableKey: VariableKey,
     options: DevCycleServerSDKOptions,
 ): Promise<Response> {
     const baseUrl = `${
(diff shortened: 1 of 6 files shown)


cdruxerman:
Hey @SimeonC thanks so much for submitting the PR, our team will take a look and let you know our thoughts.

ajwootto:
Hey @SimeonC thanks for the PR! We do currently have an option on the CLI generator where you can output re-defined versions of the `useVariableValue` hooks if you pass `--react` to the generator. This basically gives you new definitions for the hooks that you import and use rather than using the ones directly from the SDK. [More Info](https://docs.devcycle.com/sdk/client-side-sdks/react/react-typescript)

However, if I understand this correctly you're basically allowing the type definition of the existing methods to be augmented with `declare module`? This seems like it might be a better approach than the current way we do it.

SimeonC:
I won't take credit for the idea, it's a technique that [react-i18next](https://www.i18next.com/overview/typescript) uses and I thought I could adapt to here. This way of doing it also means the base js library no longer needs to pass in generics as the default is augmented correctly.

I think the biggest advantage of doing it the declare module way is that the library usage stays the same way as in the docs, we don't need to remember to import from the generated file instead of the library directly.

In this PR I've included the `SwapComponents` and `RenderIfEnabled` helpers to also use the new types too as in my screenshot.

ajwootto:
@SimeonC sounds good. I've been working on a PR based on this that also takes care of some more of our type safety guarantees, like making sure the default value and return type of `useVariableValue` are correct according to the custom definitions 

you can see that happening here: https://github.com/DevCycleHQ/js-sdks/pull/1002

should be landing next week

SimeonC:
Very nice! I'll close this one off then - thanks for picking this up and completing it! I'll watch the other PR and put in any comments if I see something. Though it looks like you've got a good handle on it 👍

PR #1002, opened 2024-11-28
feat: make variable key types overridable via module declaration

ajwootto:
Based on https://github.com/DevCycleHQ/js-sdks/pull/1001

Update the type definitions of our SDKs to support overriding the definition of allowable variables via `declare module` interface merging. This makes it so that something like this becomes fully type-safe:

```typescript
declare module '@devcycle/types' {
    interface CustomVariableDefinitions {
        flag: 'a' | 'b'
    }
}
const test = useVariableValue('flag', 'b')
// typeof test is `'a' | 'b'`

// type error here, 'c' is not a valid value for 'flag'
const test = useVariableValue('flag', 'c')

// type error here, 'unknown' is not a valid key
const test2 = useVariableValue('unknown', false)
```


ehis-042:
Nice project

kaushalkapasi (approved):

lib/shared/types/src/types/variableKeys.ts
```diff
@@ -0,0 +1,46 @@
...
+ * import '@devcycle/types';
+ * declare module '@devcycle/types' {
+ *   interface CustomVariableDefinitions extends DVCVariableTypes {}
+ * }
+ * ```
+ */
+// eslint-disable-next-line @typescript-eslint/no-empty-interface
+export interface CustomVariableDefinitions {}
+type DynamicBaseVariableDefinitions =
+    keyof CustomVariableDefinitions extends never
```
jonathannorris, line 25: How does the `never` work here? How would this statement ever be evaluated?

jonathannorris: interesting: 

```
1. keyof CustomVariableDefinitions extends never: This expression checks if the CustomVariableDefinitions interface has any keys.
- keyof CustomVariableDefinitions produces a union of the keys of the CustomVariableDefinitions interface. If the interface has no keys, this results in the never type.
- extends never checks if this union is never, which would mean that CustomVariableDefinitions has no keys.
```

SimeonC: For some more explanation even though I think you probably got the gist of it from the ChatGPT/Copilot explanation (this was my code from the previous PR);

The interface `CustomVariableDefinitions` can be augmented via a process called "Declaration Merging" which has an [entire page dedicated to it in the TS docs](https://www.typescriptlang.org/docs/handbook/declaration-merging.html) (meaning it's not all that straightforward 😆 ).

So because of this `CustomVariableDefinitions` could be either an empty object if the client code hasn't augmented it, or an object with keys if it has been augmented. Meaning the interface kind of acts like a generic - the non declaration merging way that people would usually do in their own client code would be `type DynamicBaseVariableDefinitions<TDefs> = keyof TDefs extends never ? ...`

Using `CustomVariableDefinitions` on it's own would require users to always do the declaration merging or any "variable key" argument would be defined as `key: never`.
Particularly for those non-typescript users who just use types "invisibly" as IDE autocomplete or just the "automatic get out of my way" typescript users we want the key parameters to be `string` typed.

With all that, you could read the type definition in natural language as;

```
if CustomVariableDefinitions has not been augmented by Declaration merging (ie keys are never)
then VariableDefinitions should be `Record<string, string | number | boolean>`
otherwise use the strict typed version
```

PR #1004, opened 2024-12-02
fix: fixed bug with rollout when starting with 100 instead of 0

jsalaber:
# Changes

- fixed bug when user doesn't qualify for next stage returning 0, when start percentage is non 0


lib/shared/bucketing/src/bucketing.ts
```diff
@@ -105,7 +105,7 @@ export const getCurrentRolloutPercentage = (
             : null)
 
     if (!currentStage) {
-        return 0
+        return start
```
ajwootto, line 108: pretty sure this just introduces a bug where it puts everyone in the start percentage of the rollout before the start date is actually triggered.

ajwootto (changes requested):

lib/shared/bucketing/__tests__/bucketing.test.ts
```diff
@@ -1415,6 +1415,118 @@ describe('Rollout Logic', () => {
...
+                        ),
+                        percentage: 1,
+                    },
+                ],
+            }
+
+            // Before next stage - should be 50%
+            jest.useFakeTimers().setSystemTime(new Date())
+            expect(
+                doesUserPassRollout({ rollout, boundedHash: 0.49 }),
```
ajwootto, line 1475: based on my understanding of how stepped rollouts are supposed to work, this test is wrong

jsalaber: what's wrong with it?

ajwootto: see slack

sdk/nextjs/src/common/transformConfig.ts
```diff
@@ -0,0 +1,40 @@
+import { ConfigBody } from '@devcycle/types'
+
+/**
+ * Transform the config to ensure all dates are valid Date objects
+ * @param configData
+ * @returns
+ */
+export const transformConfig = (configData: ConfigBody): ConfigBody => {
```
ajwootto, line 8: why not use class-transformer to use the transformations already defined?

sdk/nextjs/src/pages/bucketing.ts
```diff
@@ -52,7 +53,7 @@ const bucketOrFetchConfig = async (
 
     return generateBucketedConfig({
         user,
-        config,
+        config: transformConfig(config),
```
ajwootto, line 56: you should fix this at the fetching layer since the response is being asserted to be a ConfigBody, which it isn't

sdk/nextjs/src/common/transformConfig.ts
ajwootto, line 8: the input arg here is not a ConfigBody since it hasn't been transformed into one yet

lib/shared/bucketing/__tests__/bucketing.test.ts
```diff
@@ -1415,6 +1415,124 @@ describe('Rollout Logic', () => {
...
+            for (let i = 0; i < 100; i++) {
+                expect(
+                    doesUserPassRollout({ rollout, boundedHash: i / 100 }),
+                ).toBeTruthy()
+            }
+
+            jest.useRealTimers()
+        })
+
+        it.only('should handle stepped rollout with 50% start and 100% later stage', () => {
```
ajwootto, line 1458: remove it.only

lib/shared/bucketing/__tests__/bucketing.test.ts
```diff
@@ -1415,6 +1415,124 @@ describe('Rollout Logic', () => {
...
+            for (let i = 0; i < 100; i++) {
+                expect(
+                    doesUserPassRollout({ rollout, boundedHash: i / 100 }),
+                ).toBeTruthy()
+            }
+
+            jest.useRealTimers()
+        })
+
+        it('should handle stepped rollout with 100% start and 0% later stage', () => {
```
ajwootto, line 1498: can you add a test case for when the starting step is in the future, to make sure it's zero

ajwootto (approved):

PR #1012, opened 2024-12-12
feat: React Native SSE Support

jonathannorris:
Adding SSE connections to the React-Native SDK to enable live-updates of configs.
- using `react-native-sse` library
- re-factored JS SDK SSE connection to pass in `react-native-sse` Event Source class from react native lib

**Note:** will need to mention in the docs the limitations on react native <0.75 requiring flipper to be disabled: https://github.com/binaryminds/react-native-sse/issues/61


sdk/js/src/Client.ts
```diff
@@ -746,11 +751,16 @@ export class DevCycleClient<
...
         // Update the streaming connection URL if it has changed (for ex. if the current user has targeting overrides)
+        this.logger.debug(`Config SSE URL: ${config?.sse?.url}`)
         if (config?.sse?.url) {
-            // construct the streamingConnection if necessary
             if (!this.streamingConnection) {
                 if (!this.options.disableRealtimeUpdates) {
-                    this.streamingConnection = new StreamingConnection(
+                    const SSEConnectionClass =
+                        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+                        ((this.options as any)
```
ajwootto, line 760: ?

sdk/js/src/Client.ts
```diff
@@ -125,6 +128,8 @@ export class DevCycleClient<
             this.logger,
         )
 
+        console.log(`Test `)
```
ajwootto, line 131: log

sdk/react-native/src/ReactNativeSSEConnection.ts
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    public updateURL(url: string): void {
+        this.close()
+        this.url = url
+        this.openConnection()
+    }
+
+    private openConnection() {
+        this.logger.debug('ReactNativeSSEConnection opening connection, test 2')
```
ajwootto, line 23: cleanup

sdk/react-native/package.json
```diff
@@ -28,9 +28,11 @@
     "dependencies": {
         "@devcycle/js-client-sdk": "^1.32.2",
         "@devcycle/react-client-sdk": "^1.30.2",
+        "@devcycle/types": "^1.19.2",
         "@react-native-async-storage/async-storage": "^1.17.11",
         "react-native-device-info": "^8.7.0",
-        "react-native-get-random-values": "^1.7.2"
+        "react-native-get-random-values": "^1.7.2",
+        "react-native-sse": "^1.2.1"
```
suthar26, line 35: note: this will require docs update as we need to install the other packages separately

suthar26 (approved):

sdk/react-native/package.json
jonathannorris: Why wouldn't it be installed from the dependency? seems to work in the E2E test app.

suthar26: im not 100 percent sure, but we have that explicit in the docs and in the past when i have tried to install, required me to add those dependencies have not tested recently tbh

PR #1014, opened 2024-12-16
fix calling getVariableValue in server actions

ajwootto:
- Next SDK was previously relying on the ability to mutate the returned object of a specific React "cache" call, in order to set and get a value that can be used across various functions
- first of all, this pattern is apparently discouraged in general apparently, they want us to write code using "cache" that assumes the result is re-derivable at any time if the cache were suddenly cleared (ie. dont rely on the object returned from cache always being the same or always being there)
- second of all, cache doesnt work in server actions 
- so to fix using the SDK in server actions (and middleware in theory), remove the use of the mutating cached context and simply pass the client around to the places its needed
- remove the `getVariableValue` etc. methods that were exported from their own files and instead just directly call the client's methods inside the "setupDevCycle" scope


sdk/nextjs/src/server/setupDevCycle.tsx
```diff
@@ -54,7 +76,10 @@ export const setupDevCycle = <
             clientSDKKey,
             userGetter,
             options,
-        )
+        ).then((result) => {
```
jsalaber, line 79: can this function be async?

ajwootto: no, this is explicitly passing an unresolved promise

elliotCamblor (approved):

PR #1028, opened 2025-02-13
feat: create @devcycle/openfeature-angular-provider

jonathannorris:
- create OpenFeature Angular provider


JamieSinn (commented):
Is there a startup test we can do like the other providers?

jonathannorris:
@JamieSinn I don't believe we have startup tests for the OpenFeature providers. And this is essentially a one line change off the web provider.

JamieSinn (approved):

PR #1029, opened 2025-02-24
feat: change OpenFeature imports and DevCycleProvider interface

jonathannorris:
- Change OpenFeature imports from a dynamic import (which has always caused us problems), to a straight require.
- This is a bit of a reversion to the old way we had it setup, and will require all SDK users using Yarn to add the OpenFeature packages.
- But this will mean that all SDKs will come with OpenFeature setup (good thing!), and the `getOpenFeatureProvider()` method no longer needs to be `async` which was causing complications with SDKs like NestJS.

### DIFF
--- a/dev-apps/openfeature-nodejs/src/main.ts
+++ b/dev-apps/openfeature-nodejs/src/main.ts
@@ -65,7 +65,7 @@ async function startDevCycle() {
         ...context,
         user_id: context.targetingKey,
     }
-    const allFeatures = await provider.devcycleClient.allFeatures(dvcUser)
+    const allFeatures = provider.devcycleClient.allFeatures(dvcUser)
     console.log(`All Features: ${JSON.stringify(allFeatures)}`)
 }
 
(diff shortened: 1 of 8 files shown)


dev-apps/openfeature-nodejs/src/main.ts
```diff
@@ -1,17 +1,16 @@
...
 
 let openFeatureClient: Client
 
 async function startDevCycle() {
-    await OpenFeature.setProviderAndWait(
-        await initializeDevCycle(
-            DEVCYCLE_SERVER_SDK_KEY,
-        ).getOpenFeatureProvider(),
-    )
+    const provider = new DevCycleProvider(DEVCYCLE_SERVER_SDK_KEY, {
```
jonathannorris, line 10: Example of how the initialization changes.

dev-apps/openfeature-nodejs/src/main.ts
```diff
@@ -61,6 +60,22 @@ async function startDevCycle() {
         metaDataFieldBoolean: true,
         metaDataFieldDouble: 1.23,
     })
+
+    const dvcUser = {
+        ...context,
+        user_id: context.targetingKey,
+    }
+    console.log(
```
kaushalkapasi, line 68: can we remove these since these are debug functions and instead highlight our `variable` or `variableValue` calls instead

jonathannorris: going to leave the `allFeatures()` call as I want an example of how to access the `devcycleClient` from the provider.

kaushalkapasi (approved):

PR #1036, opened 2025-03-19
feat: add setClientCustomData to the local nodejs server sdk

jonathannorris:

### DIFF
--- a/sdk/nodejs/src/client.ts
+++ b/sdk/nodejs/src/client.ts
@@ -457,10 +457,12 @@ export class DevCycleClient<
         clearInterval(this.bucketingTracker)
     }
 
-    async setClientCustomData(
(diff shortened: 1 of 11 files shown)


JamieSinn (approved):
I forgot how easy this is with WASM

sdk/nodejs/src/client.ts
```diff
@@ -455,4 +456,16 @@ export class DevCycleClient<
         this.eventQueue.cleanup()
         clearInterval(this.bucketingTracker)
     }
+
+    setClientCustomData(clientCustomData: DVCCustomDataJSON): void {
```
jonathannorris, line 460: @JamieSinn changing this to be sync like all the other SDKs to work with the test harness

JamieSinn: kk

PR #1039, opened 2025-03-24
chore: update yarn to 4.6.0, update deps

jonathannorris:
- Update to yarn `4.6.0` everywhere using corepack to install, remove the stored yarn versions.
- Updated dependencies using `yarn upgrade-interactive`
- Fixed new TS issues


JamieSinn (approved):
should probably do this for other repos too

kaushalkapasi (approved):

PR #1040, opened 2025-03-25
chore: add option to not throw error on user error in dvc js and react

jsalaber:
# Changes

- added option to js sdk to not throw `UserError` if the user specifies option

### DIFF
--- a/sdk/js/src/index.ts
+++ b/sdk/js/src/index.ts
@@ -105,7 +105,7 @@ export function initializeDevCycle<
         | DevCycleOptionsWithDeferredInitialization,
     optionsArg: DevCycleOptions = {},
 ): DevCycleClient<Variables, CustomData> {
-    if (!sdkKey) {
+    if (!sdkKey && optionsArg?.throwOnError) {
         throw new UserError(
             'Missing SDK key! Call initialize with a valid SDK key',
         )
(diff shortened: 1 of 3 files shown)


sdk/js/src/index.ts
```diff
@@ -105,7 +105,7 @@ export function initializeDevCycle<
         | DevCycleOptionsWithDeferredInitialization,
     optionsArg: DevCycleOptions = {},
 ): DevCycleClient<Variables, CustomData> {
-    if (!sdkKey) {
+    if (!sdkKey && optionsArg?.throwOnError) {
```
kaushalkapasi, line 108: can we change this to have an IF statement inside which will at least `this.logger.error` instead of failing silently?

Also the `throwOnError` should default to true as that is our existing behaviour.
 
```suggestion
    if (!sdkKey) {
      if (optionsArg?.throwOnError) {
        throw new UserError(
            'Missing SDK key! Call initialize with a valid SDK key',
        )
      } else {
        this.logger.error('Missing SDK key! Call initialize with a valid SDK key)
      }
    }
```

sdk/react/src/DevCycleProvider.tsx
```diff
@@ -19,13 +19,15 @@ export function DevCycleProvider(props: Props): React.ReactElement {
...
+    const throwOnError =
+        options?.throwOnError !== undefined ? options.throwOnError : true
     let sdkKey: string
     if ('sdkKey' in config) {
         sdkKey = config.sdkKey
     } else {
         sdkKey = config.envKey
     }
-    if (!sdkKey) {
+    if (!sdkKey && !options?.throwOnError) {
```
kaushalkapasi, line 30: similar to previous, split this into a nested IF which will log an error if the key is missing but `throwOnError` is false

PR #1043, opened 2025-04-10
feat: switch to @protobuf-ts lib from protobufjs

jonathannorris:
- switch to @protobuf-ts from protobufjs, better TS support and no eval use


sdk/nodejs/src/utils/userBucketingHelper.ts
```diff
@@ -75,15 +74,12 @@ export function variableForUser_PB(
         variableType: type,
         shouldTrackEvent: true,
     }
-    const err = VariableForUserParams_PB.verify(params)
```
jonathannorris, line 78: TODO: confirm verify behaviour isn't needed.

PR #1046, opened 2025-04-25
chore: add _feature to SDKVariable on clientSDKAPI

jonathannorris:
- adds the `_feature` id to the Variables returned by the SDK API. This will be used on the client SDKs to know how to filter their `featureVars` for each `variableEvaluation` / `variableDefault` event.
- updates the `BucketedUserConfig` schema with an optional `settings.filterFeatureVars` boolean to signal to the SDKs that `_feature` params should exist on Variables in the config and they should filter `featureVars`. This is to let SDKs filter `featureVars ` to an empty object if there isn't a `_feature` for a variable.

### DIFF
--- a/lib/shared/types/src/types/apis/sdk/clientSDKAPI.ts
+++ b/lib/shared/types/src/types/apis/sdk/clientSDKAPI.ts
@@ -343,6 +343,7 @@ export class DVCOptInUser {
 
 export type SDKVariable = PublicVariable & {
     value: VariableValue
+    _feature?: string
     evalReason?: unknown
 }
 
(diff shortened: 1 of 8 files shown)


kaushalkapasi (approved):

lib/shared/types/src/types/apis/sdk/clientSDKAPI.ts
```diff
@@ -413,4 +414,14 @@ export interface BucketedUserConfig {
...
+
+    /**
+     * Settings for the bucketed config.
+     */
+    settings?: {
+        /**
+         * If true, the bucketed config will filter out featureVars per evaluation.
+         */
+        filterFeatureVars?: boolean
+    }
```
JamieSinn, line 426: Does this exist on the worker model yet?

JamieSinn (approved):

PR #1049, opened 2025-04-29
feat: add _feature to variable response of bucketing

jonathannorris:
- Add `_feature` param to `variables` response of bucketing lib and AS bucketing lib
- Will be used by the SDK API responses to the client SDKs


JamieSinn (approved):
Are we going to extend this to the other libs?

jonathannorris:
> Are we going to extend this to the other libs?

We can add it to GO if we want, but its not really needed on the server-side because we do agg events there.

PR #1055, opened 2025-05-14
chore: test ubicloud actions runner

jonathannorris:


.github/workflows/nx-affected-test.yml
```diff
@@ -21,11 +23,11 @@ jobs:
                   cache: 'yarn'
             - run: git fetch origin main
             - run: yarn --immutable
-            - run: yarn affected:lint --parallel=4
+            - run: yarn affected:lint --parallel=8
```
JamieSinn, line 26: at some point I'd add this as a matrix option so that we can just change it in one location similar to how we have the node version defined.

.github/workflows/nx-affected-e2e.yml
```diff
@@ -4,7 +4,7 @@ on:
         branches: [main]
 jobs:
     build:
-        runs-on: ubuntu-latest
+        runs-on: ubicloud-standard-8
```
JamieSinn, line 7: can we run anything here parallel with the playwright params?

JamieSinn (approved):

PR #1061, opened 2025-05-16
fix: check all filters on top level ORs for a truthy match before returning false

kaushalkapasi:
- fix: check all filters defined when the `filter.operator` is an OR before returning true or false

### DIFF
--- a/lib/shared/bucketing-assembly-script/assembly/bucketing/segmentation.ts
+++ b/lib/shared/bucketing-assembly-script/assembly/bucketing/segmentation.ts
@@ -36,8 +36,10 @@ export function _evaluateOperator(
         for (let i = 0; i < operator.filters.length; i++) {
             const filter = operator.filters[i]
             if (filter.operatorClass !== null) {
-                const evalResult = _evaluateOperator(filter.operatorClass as AudienceOperator, audiences, user, clientCustomData)
-                // Instead of returning the value from only the first filter, we want to return true if any of the filters are true
+                const evalResult =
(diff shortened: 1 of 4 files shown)


lib/shared/bucketing-assembly-script/assembly/bucketing/segmentation.ts
```diff
@@ -36,7 +36,11 @@ export function _evaluateOperator(
         for (let i = 0; i < operator.filters.length; i++) {
             const filter = operator.filters[i]
             if (filter.operatorClass !== null) {
-                return _evaluateOperator(filter.operatorClass as AudienceOperator, audiences, user, clientCustomData)
+                const evalResult = _evaluateOperator(filter.operatorClass as AudienceOperator, audiences, user, clientCustomData)
+                // Instead of returning the value from only the first filter, we want to return true if any of the filters are true
+                if (evalResult) {
```
suthar26, line 41: this makes sense

lib/shared/bucketing-test-data/src/data/testData.ts
```diff
@@ -227,6 +298,34 @@ export const audiences: TargetAudience[] = [
...
     },
+    {
+        _id: '',
+        filters: {
+            filters: [
+                {
+                    type: FilterType.audienceMatch,
+                    comparator: FilterComparator['='],
+                    // Top Level OR audience
+                    _audiences: ['614ef6ea475929459060721b'],
```
suthar26, line 309: should we use the `reusableTopLevelOrAudience._id` here?

suthar26 (approved):

PR #1067, opened 2025-05-26
fix: all non next.js /dev-apps

jonathannorris:


dev-apps/nextjs/app-router/app/layout.tsx
```diff
@@ -1,20 +1,16 @@
...
     children: React.ReactNode
 }) {
     return (
         <html lang="en">
-            <body>
-                <DevCycleClientsideProvider context={getClientContext()}>
-                    {children}
-                </DevCycleClientsideProvider>
-            </body>
+            <body>{children}</body>
```
phaym, line 13: this was happening to me too, next generatesa layout even though we have one. its not right because the dvc provider is no longer there.  I didn't figure out why

phaym (approved):

PR #1073, opened 2025-05-28
fix: missing document check in close

jonathannorris:
## Summary
- update `close` to check for `document` before removing event listener

## Testing
- `nx test js` *(fails: request to https://registry.npmjs.org/nx failed)*


jonathannorris:
From OpenAI Codex

PR #1085, opened 2025-06-03
feat: add eval reasons for JS Bucketing

kaushalkapasi:
- feat: adds Evaluation Reasons for JS Bucketing shared library as `eval` on the Feature & Variable
- chores:
  - updates tests to verify proper eval reasons are reported
  - update test data (filters) to properly use Country as 2 letter codes (ISO-3166)
  - remove unsupported IP address filter
  - only return `reasonDetails` if the evaluation is truthy
  - streamline Eval Reason Details into an Enum
- fix: update JS & NodeJS SDKs to properly parse `variable.eval` from Bucketed User Configurations
- chore: update OpenFeature Providers for DevCycle SDKs to map `variable.eval?.reason` to `reason` for evaluation reasons (aka if DevCycle provides an evaluation reason, pass it along to the OpenFeature Provider)
- fix: re-adds the `evalReason` field to the `SDKVariable` type definition to allow NodeJS builds to work, this will be removed during the WASM Bucketing lib changes

### DIFF
--- a/lib/shared/types/src/types/apis/sdk/clientSDKAPI.ts
+++ b/lib/shared/types/src/types/apis/sdk/clientSDKAPI.ts
@@ -27,18 +27,6 @@ export type SDKTypes = (typeof SDKTypeValues)[number]
 
 export type QueryParams = { [key: string]: string }
 
-export enum DEFAULT_REASONS {
-    MISSING_CONFIG = 'MISSING_CONFIG',
-    MISSING_VARIABLE = 'MISSING_VARIABLE',
-    MISSING_FEATURE = 'MISSING_FEATURE',
-    MISSING_VARIATION = 'MISSING_VARIATION',
-    MISSING_VARIABLE_FOR_VARIATION = 'MISSING_VARIABLE_FOR_VARIATION',
-    USER_NOT_IN_ROLLOUT = 'USER_NOT_IN_ROLLOUT',
-    USER_NOT_TARGETED = 'USER_NOT_TARGETED',
-    INVALID_VARIABLE_TYPE = 'INVALID_VARIABLE_TYPE',
-    UNKNOWN = 'UNKNOWN',
-}
-
 export enum EVAL_REASONS {
     TARGETING_MATCH = 'TARGETING_MATCH',
     SPLIT = 'SPLIT',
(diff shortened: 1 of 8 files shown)


lib/shared/bucketing/src/segmentation.ts
```diff
@@ -112,39 +154,97 @@ function filterForAudienceMatch({
...
     customData: (data, filter) => {
         const combinedCustomData = {
             ...data.customData,
             ...data.privateCustomData,
         }
-        return checkCustomData(combinedCustomData, filter)
+        const result = checkCustomData(combinedCustomData, filter)
+        return {
+            result,
+            reasonDetails: result ? `Custom Data ${filter.dataKey}` : undefined,
```
JamieSinn, line 246: i would make this also generically apply to say "custom data or private custom data"

lib/shared/bucketing/src/segmentation.ts
```diff
@@ -112,39 +154,97 @@ function filterForAudienceMatch({
...
+            reasonDetails: result ? `Email` : undefined,
+        }
+    },
+    ip: (data, filter) => {
+        const result = checkStringsFilter(data.ip, filter)
+        return {
+            result,
+            reasonDetails: result ? 'IP' : undefined,
+        }
+    },
```
JamieSinn, line 202: we don't even have this filter? wat.

JamieSinn: holdover from TL - I'd kill it

lib/shared/bucketing/src/segmentation.ts
```diff
@@ -73,20 +91,44 @@ export const evaluateOperator = ({
...
+
+        return result
     }
 
     if (operator.operator === 'or') {
-        return operator.filters.some(doesUserPassFilter)
+        const result = operator.filters.some(doesUserPassFilter)
+        return {
+            result,
+            reasonDetails: result ? reason : '',
```
JamieSinn, line 116: why emptystring instead of undefined like the other returns?

lib/shared/bucketing/src/segmentation.ts
kaushalkapasi: From a filter perspective, its just a Custom Data match

lib/shared/bucketing/src/segmentation.ts
```diff
@@ -112,39 +163,96 @@ function filterForAudienceMatch({
...
-    appVersion: (data, filter) => checkVersionFilters(data.appVersion, filter),
-    platformVersion: (data, filter) =>
-        checkVersionFilters(data.platformVersion, filter),
-    deviceModel: (data, filter) => checkStringsFilter(data.deviceModel, filter),
-    platform: (data, filter) => checkStringsFilter(data.platform, filter),
+    country: (data, filter) => {
+        const result = checkStringsFilter(data.country, filter)
+        return {
+            result,
+            reasonDetails: result ? EVAL_REASON_DETAILS.COUNTRY : undefined,
```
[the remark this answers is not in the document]
kaushalkapasi: if result is `false`, that means the user did not meet the criteria to be bucketed into the feature, therefore we don't want to set the `reasonDetails` property

JamieSinn (approved):

JamieSinn (approved):

lib/shared/bucketing/src/segmentation.ts
```diff
@@ -112,39 +163,96 @@ function filterForAudienceMatch({
...
-    email: (data, filter) => checkStringsFilter(data.email, filter),
-    ip: (data, filter) => checkStringsFilter(data.ip, filter),
-    user_id: (data, filter) => checkStringsFilter(data.user_id, filter),
-    appVersion: (data, filter) => checkVersionFilters(data.appVersion, filter),
-    platformVersion: (data, filter) =>
-        checkVersionFilters(data.platformVersion, filter),
-    deviceModel: (data, filter) => checkStringsFilter(data.deviceModel, filter),
-    platform: (data, filter) => checkStringsFilter(data.platform, filter),
+    country: (data, filter) => {
+        const result = checkStringsFilter(data.country, filter)
```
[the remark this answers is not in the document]
suthar26: or is the idea to let the user know of the field that they are getting evaluated by and not the actual targeting?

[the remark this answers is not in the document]
kaushalkapasi: the field(s) that the user matched to get bucketed into a given variation

PR #1087, opened 2025-06-05
fix: handle the bounded hash limit for edge cases

suthar26:
- add checking for rare case when the bounded hash is 1 
- add test for the bucketing libs


lib/shared/bucketing-assembly-script/assembly/types/targetV2.ts
```diff
@@ -58,7 +58,7 @@ export class TargetV2 extends JSON.Value {
         for (let i = 0; i < this._sortedDistribution.length; i++) {
             const distribution = this._sortedDistribution[i]
             distributionIndex += distribution.percentage
-            if (boundedHash >= previousDistributionIndex && boundedHash < distributionIndex) {
+            if (boundedHash >= previousDistributionIndex && (boundedHash < distributionIndex || (distributionIndex == 1 && boundedHash == 1))) {
```
kaushalkapasi, line 61: @suthar26 fix this lint warning, fails the build

PR #1089, opened 2025-06-09
chore: add tests for invalid config response

phaym:
it looks like the js-sdk handles errors on initializing so i just added these tests to confirm the same scenario that caused issue in ios sdk is covered. Please suggest any other tests/dependant libraries i could add tests to but i think they rely directly on the js initialization.

### DIFF
--- a/sdk/js/__tests__/index.spec.js
+++ b/sdk/js/__tests__/index.spec.js
@@ -96,8 +96,9 @@ describe('initializeDevCycle tests', () => {
     it('should not throw an error and use default config when config api throws an error', async () => {
         const user = { user_id: 'testuser' }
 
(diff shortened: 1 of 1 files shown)


sdk/js/__tests__/index.spec.js
```diff
@@ -18,6 +15,23 @@ const invalidOptionsError =
     'Invalid options! Call initialize with valid options'
 
 describe('initializeDevCycle tests', () => {
+    /**
+     * @type {jest.SpyInstance}
+     */
```
jonathannorris, line 20: This probably isn't needed.

```suggestion
```

phaym: i like to type this so i have the intellisense on the functions available to me, why don't you think its needed?

PR #1090, opened 2025-06-11
feat: added eval hooks for js

jsalaber:
# Changes

- added EvalHooks to js sdk
- move variable logic to internal method so that hooks runner can wrap around variable method

### DIFF
--- a/sdk/js/__tests__/Client.test.ts
+++ b/sdk/js/__tests__/Client.test.ts
@@ -96,4 +96,51 @@ describe('DevCycleClient', () => {
             },
         })
     })
+
+    it('should call hooks when variable is evaluated', () => {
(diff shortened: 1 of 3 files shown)


lib/shared/hooks/src/evalHooksRunner.ts
```diff
@@ -0,0 +1,61 @@
...
+        resolver: () => any,
+    ) {
+        const context = new HookContext(user, key, {})
+        const savedHooks = [...this.hooks]
+        const reversedHooks = this.hooks.reverse()
+
+        let variable: any
+        try {
+            await this.runBefore(savedHooks, context)
+            variable = await resolver.call(context)
```
jsalaber, line 19: call resolver with context saved from runBefore

suthar26 (approved):

PR #1096, opened 2025-06-17
feat: update identifyUser logic to use cached user config if exists

jonathannorris:
## Improve error handling for identifyUser and resetUser operations

This PR enhances the error handling behavior for `identifyUser` and `resetUser` methods in the JavaScript SDK to provide better resilience and user experience.

### Changes

- **Enhanced `identifyUser` error handling:**
  - Falls back to cached user config when config fetch fails (with warning log)
  - Throws error without changing client state when no cached config is available
  - Properly handles null config responses from server

- **Improved `resetUser` error handling:**
  - Restores original anonymous ID when config fetch fails
  - Maintains client state integrity during error conditions
  - Better rollback mechanism for failed operations

- **Added null config protection:**
  - Added safety check in `handleConfigReceived` to prevent null reference errors
  - Proper error logging when config fetch returns null

- **Comprehensive test coverage:**
  - Added test cases for all new error handling scenarios
  - Tests verify client state remains unchanged on failures
  - Tests confirm proper fallback to cached configurations

### Benefits

- More resilient SDK behavior during network issues
- Better offline/cached config support
- Improved error messages and logging
- Maintains client state consistency during failures


sdk/js/__tests__/Client.spec.js
```diff
@@ -741,6 +741,81 @@ describe('DevCycleClient tests', () => {
             expect(anonUserId).toBe(originalAnonUserId)
             expect(anonUserId).toBeDefined()
         })
+
+        it('should throw error when config fetch returns null', async () => {
+            const newUser = { user_id: 'user2' }
+            getConfigJson_mock.mockClear()
+            getConfigJson_mock.mockResolvedValue(null)
+
+            await expect(client.identifyUser(newUser)).rejects.toThrow(
```
suthar26, line 750: can we add a client.variable() call to ensure we still get default values here

suthar26 (approved):

PR #1097, opened 2025-06-18
feat: add eval reasons to WASM Bucketing

phaym:
Add "Eval Reasons" to Variables and Features to report what condition trigger a variation to be served, similar to changes for js library in [#1085](https://github.com/DevCycleHQ/js-sdks/pull/1085)
- updated tests to include the expected eval reason, details and target_id

Closes COR-3426


lib/shared/bucketing-assembly-script/__tests__/bucketing/segmentation.test.ts
```diff
@@ -1215,49 +1296,76 @@ describe('SegmentationManager Unit Test', () => {
...
+            )
+            assert.deepStrictEqual(
+                { result: false },
+                checkStringsFilter('', filter),
+            )
         })
         it('should return true if exists filter and value', () => {
             const filter = { type: 'user', comparator: 'exist' }
-            assert.strictEqual(true, checkStringsFilter('string', filter))
+            assert.deepStrictEqual(
```
phaym, line 1327: anything using checkStringsFilter is returning 'Email' as a reason detail, because there is logic in that function adding a email filter that im hesitant to change right now.

.github/workflows/run-test-harness.yml
```diff
@@ -36,3 +36,4 @@ jobs:
                   sdks-to-test: nodejs,of-nodejs
                   sdk-github-sha: ${{github.event.pull_request.head.sha}}
                   github-token: ${{ secrets.AUTOMATION_USER_TOKEN }}
+                    sdk-capabilities: '{ "NodeJS":["edgeDB", "cloud", "lastModifiedHeader", "bootstrapping", "sdkConfigEvent", "clientUUID", "v2Config", "clientCustomData", "variablesFeatureId", "cloudEvalReason", "evalReason"], "OF_NodeJS": ["edgeDB", "cloud", "lastModifiedHeader", "sdkConfigEvent", "clientUUID", "v2Config", "sdkPlatform", "variablesFeatureId","evalReason"]}'
```
kaushalkapasi, line 39: this change should be on the test-harness, not in this workflow

lib/shared/bucketing-assembly-script/__tests__/protobufVariableHelper.ts
```diff
@@ -9,6 +14,11 @@ type SDKVariable_PB_Type = {
     boolValue: boolean
     doubleValue: number
     stringValue: string
+    eval?: {
+        reason: EVAL_REASONS
+        details?: string
+        targetId?: string
```
kaushalkapasi, line 20: mismatch, should be `target_id`

phaym: k updated, and updated test_harness so pr will pass https://github.com/DevCycleHQ/test-harness/pull/533

lib/shared/bucketing-assembly-script/assembly/types/target.ts
```diff
@@ -50,15 +78,19 @@ export class Target extends JSON.Value {
...
         const previousDistributionIndex: f64 = 0
+        const isRollout = this.rollout !== null
+        const isRandomDistribution = this.distribution.length !== 1
         for (let i = 0; i < this._sortedDistribution.length; i++) {
             const distribution = this._sortedDistribution[i]
             distributionIndex += distribution.percentage
             if (boundedHash >= previousDistributionIndex && 
                 (boundedHash < distributionIndex || (distributionIndex == 1 && boundedHash == 1))) {
-                return distribution._variation
+                const reason = isRollout || isRandomDistribution ? EVAL_REASONS.SPLIT : EVAL_REASONS.TARGETING_MATCH
```
suthar26, line 91: why are we not adding the detail or variation to this object?

phaym: originally we didn't have target_id and split reasons in scope, i could add that here. but its a good callout, i may have neglected this because i wasn't sure in what context it is used. does this get exposed directly in sdks?

phaym: thanks for pointing this out, i didn't need to update this part of the code at all since its only responsible for returning the variation id, i followed some changes in js bucketing that led me down the wrong path

.github/workflows/run-test-harness.yml
phaym: test harness wasn't running properly due to syntax error, im now seeing slightly different failure result then i'm expecting. will see whats going wrong there

lib/shared/bucketing-assembly-script/assembly/bucketing/segmentation.ts
```diff
@@ -71,12 +102,13 @@ function doesUserPassFilter(
...
-): bool {
+): SegmentationResult {
     let isValid = true
 
-    if (filter.type === 'all') return true
-    else if (filter.type === 'optIn') return false
-    else if (filter.type === 'audienceMatch') {
+    if (filter.type === 'all') return new SegmentationResult(true, EVAL_REASON_DETAILS.ALL_USERS)
+    else if (filter.type === 'optIn') {
+            return new SegmentationResult(false)
```
suthar26, line 110: we have the eval reason for optin we should use here if you are using it for the line above

lib/shared/bucketing-assembly-script/assembly/bucketing/bucketing.ts
```diff
@@ -321,8 +360,8 @@ export function _generateBucketedConfig(
                 variable.type,
                 variable.key,
                 variationVar.value,
-                null,
```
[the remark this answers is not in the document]
suthar26: in that case why would this not be removed from the constructor?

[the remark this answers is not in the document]
suthar26: appologies, i confused this with the PB constructor, this is good to go. the constructor in buckedUserConfig.ts has been changed to remove this field

lib/shared/bucketing-assembly-script/__tests__/bucketing/bucketing.test.ts
```diff
@@ -25,15 +25,28 @@ import {
...
 } from '../protobufVariableHelper'
 
 type BoundedHash = { rolloutHash: number; bucketingHash: number }
 
+type EvalReason = {
+    reason: string
+    details?: string
+}
+
+type VariationReasonResult = {
```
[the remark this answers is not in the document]
phaym: nope this is leftover, thanks will remove

lib/shared/bucketing-assembly-script/assembly/bucketing/segmentation.ts
kaushalkapasi: The targeting rule will still exist in the config, however there is currently no use-case where it would evaluate to true, so the `Opt-In`, `Not Opt-In` and `Overrides` reasons / logic can be removed.

phaym: > Opt In can be setup as an Opt Out as well

 what do you mean by this? is the case you're covering when they are in turso but set to false because they have enabled then disabled?

lib/shared/bucketing-assembly-script/assembly/bucketing/bucketing.ts
```diff
@@ -241,9 +259,26 @@ export function bucketUserForVariation(
...
+        reason = EVAL_REASONS.SPLIT
+        const evalReasonPrefix =
+            hasMultipleDistributions && hasRollout
+                ? `${EVAL_REASON_DETAILS.RANDOM_DISTRIBUTION} | ${EVAL_REASON_DETAILS.ROLLOUT}`
+                : hasMultipleDistributions
+                ? EVAL_REASON_DETAILS.RANDOM_DISTRIBUTION
+                : EVAL_REASON_DETAILS.ROLLOUT
+        reasonDetails = `${evalReasonPrefix} | ${reasonDetails}`
+    }
+    return  new EvalReason(reason, reasonDetails, target._id)
```
JamieSinn, line 281: Can we do this without creating a new allocation here?

lib/shared/bucketing-assembly-script/assembly/bucketing/segmentation.ts
```diff
@@ -115,44 +147,61 @@ function filterForAudienceMatch(
...
+        const result = _checkVersionFilters(user.platformVersion, filter)
+        return new SegmentationResult(result, result ? EVAL_REASON_DETAILS.PLATFORM_VERSION : null)
     } else if (subType === 'deviceModel') {
-        return _checkStringsFilter(user.deviceModel, filter)
+        const result = _checkStringsFilter(user.deviceModel, filter)
+        return new SegmentationResult(result, result ? EVAL_REASON_DETAILS.DEVICE_MODEL : null)
     } else if (subType === 'platform') {
-        return _checkStringsFilter(user.platform, filter)
+        const result = _checkStringsFilter(user.platform, filter)
+        return new SegmentationResult(result, result ? EVAL_REASON_DETAILS.PLATFORM : null)
```
JamieSinn, line 195: yeah - not a huge fan of this now returning a full object/class instead of just a boolean. This is a big alloc since it's done per feature/variable when generating a full bucketed config

lib/shared/bucketing-assembly-script/__tests__/bucketing/segmentation.test.ts
phaym: https://github.com/DevCycleHQ/js-sdks/blob/main/lib/shared/bucketing-assembly-script/__tests__/bucketing/segmentation.test.ts#L54

lib/shared/bucketing-assembly-script/assembly/bucketing/bucketing.ts
jonathannorris: There are going to be new allocations for this feature, we can discuss if we want to enable folks to disable eval reasons being generated through SDK options, but I don't forsee another allocation being that large of a performance difference. Let's see what the benchmark differences look like once we merge this: https://github.com/DevCycleHQ/js-sdks/tree/main/lib/shared/bucketing-assembly-script#benchmarking

lib/shared/bucketing-assembly-script/__tests__/bucketing/bucketing.test.ts
```diff
@@ -25,15 +25,28 @@ import {
...
 import { cleanupSDK, initSDK } from '../setPlatformData'
+
 import {
     variableForUserPB,
     VariableForUserArgs,
 } from '../protobufVariableHelper'
 
 type BoundedHash = { rolloutHash: number; bucketingHash: number }
 
+type EvalReason = {
```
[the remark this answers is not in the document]
phaym: this is getting removed, i didnt need this it turned out

lib/shared/bucketing-assembly-script/assembly/bucketing/bucketing.ts
jonathannorris: Actually, @phaym can you run those benchmarks from main and this branch so we can see how they affect the performance? Make sure you rebuild the lib each time.

lib/shared/bucketing-assembly-script/assembly/types/bucketedUserConfig.ts
```diff
@@ -12,9 +12,97 @@ import {
...
+        if(!reason){
+            reason = ""
+        }
+        let details = getStringFromJSONOptional(jsonObj, 'details')
+        if(!details)
+            details = ""
+
+        let targetId = getStringFromJSONOptional(jsonObj, 'targetId')
+        if(!targetId)
+            targetId = ""
```
jonathannorris, line 93:
```suggestion
        let reason = getStringFromJSONOptional(jsonObj, 'reason')
        if (!reason) {
            reason = ""
        }
        
        let details = getStringFromJSONOptional(jsonObj, 'details')
        if (!details) {
            details = ""
        }
        
        let targetId = getStringFromJSONOptional(jsonObj, 'targetId')
        if (!targetId) {
            targetId = ""
        }
```

JamieSinn: doesn't linting take care of this?

lib/shared/bucketing-assembly-script/assembly/bucketing/segmentation.ts
```diff
@@ -9,11 +9,31 @@ import {
...
+        super()
+        this.result = result
+        this.reasonDetails = reasonDetails
+    }
+
+    stringify(): string {
+        const json = new JSON.Obj()
+        json.set('result', this.result)
+        if (this.reasonDetails !== null) {
+            json.set('reasonDetails', this.reasonDetails!)
```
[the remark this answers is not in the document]
phaym: it does serialize properly as validated in this test for example: https://github.com/DevCycleHQ/js-sdks/pull/1097/files#diff-3a51846f37ec7e08fac3a93531c681196f1385ea1e2fa88d19ca69862a53287aR357 

I just did it this way to keep it in line with what the js bucketing library did, tho i'd slightly prefer setting as empty string in both : https://github.com/DevCycleHQ/js-sdks/pull/1085/files#diff-659ae9b3e46a30142e5521775a565b33af28b15402b95585dcd2c5c57e68a79aR215

lib/shared/bucketing-assembly-script/assembly/bucketing/segmentation.ts
phaym: @kaushalkapasi  @suthar26 resolving this convo as i don't think it applies to wasm bucketing to include a reason here

lib/shared/bucketing-assembly-script/__tests__/bucketing/segmentation.test.ts
jonathannorris: could we not pass the `reason` optionally into `checkStringsFilter`?

lib/shared/bucketing-assembly-script/assembly/bucketing/segmentation.ts
phaym: i didn't find any data supporting these allocations causing memory or perf issues

lib/shared/bucketing-assembly-script/assembly/bucketing/bucketing.ts
phaym: image

the benchmark tests did show an uptick in time across the board, but i found this to be almost entirely from serializing more data and not from allocations. `variableForUser_PB` function increase from 0.019 to 0.022 which is an increase of 15.79% for each evalution.  

There is a greater perf increase in the other benchmarks then 15% but the json serialized wasm functions are not used, and generatedBucketConfig not called as frequently

lib/shared/bucketing-assembly-script/__tests__/bucketing/segmentation.test.ts
phaym: ok when i first saw this i didn't understand the intention of the test so thought it needed to change. on second look its apparently using email to test all the operations on strings. so the logic and expectations are fine this way.

suthar26 (approved):

PR #1099, opened 2025-06-18
feat: add targetId to eval, add Rollout and Random Distribution details 

phaym:


lib/shared/bucketing-assembly-script/__tests__/protobufVariableHelper.ts
```diff
@@ -17,6 +17,7 @@ type SDKVariable_PB_Type = {
     eval?: {
         reason: EVAL_REASONS
         details?: string
+        targetId?: string
```
kaushalkapasi, line 20: should be `target_id` per the spec?

PR #1106, opened 2025-06-27
feat: add _feature to the evaluated variable

suthar26:
- add _feature property to protobuf version and also to the DVC variable  

benchmark changes
image

### DIFF
--- a/sdk/js/src/Variable.ts
+++ b/sdk/js/src/Variable.ts
@@ -7,7 +7,7 @@ export interface DVCVariableOptions<T> {
     defaultValue: T
     value?: VariableTypeAlias<T>
     eval?: EvalReason
-    _feature?: string
+    _feature?: string | null
 }
 
 export class DVCVariable<T extends DVCVariableValue> implements Variable<T> {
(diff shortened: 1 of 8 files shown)


sdk/js/src/Variable.ts
```diff
@@ -34,6 +36,7 @@ export class DVCVariable<T extends DVCVariableValue> implements Variable<T> {
 
         this.defaultValue = variable.defaultValue
         this.eval = variable.eval
+        this._feature = variable._feature || undefined
```
JamieSinn, line 39: why not set it as null? that way you can just explicitly accept the value in the PB stuff?

suthar26: idk if this is related to PB, since that used SDKVariable type and that is the same type in the BucketedUserConfig. I can change it to null without any issues tho

sdk/openfeature-react-provider/__tests__/DevCycleReactProvider.test.ts
```diff
@@ -43,6 +43,7 @@ async function initOFClient(): Promise<{
                 value: true,
                 defaultValue: false,
                 isDefaulted: false,
+                _feature: 'boolean-flag',
```
JamieSinn, line 46: feature id?

sdk/openfeature-web-provider/__tests__/DevCycleProvider.test.ts
```diff
@@ -43,6 +43,7 @@ async function initOFClient(): Promise<{
                 value: true,
                 defaultValue: false,
                 isDefaulted: false,
+                _feature: 'boolean-flag',
```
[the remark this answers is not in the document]
suthar26: yeah it should be feature id, fixed it.

JamieSinn (approved):

PR #1107, opened 2025-07-02
feat: added variable definition typing for decorators in nestjs sdk

jsalaber:
# Changes

- changed variable decorators to use variable definitions so that there's more strict typing on the variable keys if the user chooses to define them

### DIFF
--- a/sdk/nestjs/src/VariableValue.ts
+++ b/sdk/nestjs/src/VariableValue.ts
@@ -5,23 +5,31 @@ import {
     DevCycleUser,
 } from '@devcycle/nodejs-server-sdk'
 import { ClsServiceManager } from 'nestjs-cls'
+import { VariableDefinitions } from '@devcycle/types'
 
-type VariableValueData = {
-    key: string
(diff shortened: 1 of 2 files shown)


sdk/nestjs/src/VariableValue.ts
```diff
@@ -5,23 +5,31 @@ import {
     DevCycleUser,
 } from '@devcycle/nodejs-server-sdk'
 import { ClsServiceManager } from 'nestjs-cls'
+import { VariableDefinitions } from '@devcycle/types'
 
-type VariableValueData = {
-    key: string
-    default: DVCVariableValue
+export type VariableValueData<
```
kaushalkapasi, line 10: does this appropriately handle for new keys that are added? do we need to add or update any tests to valdiate?

jsalaber: I'm not sure how we'd make a test for that

kaushalkapasi (approved):

PR #1114, opened 2025-07-16
fix: tighten package.json dependancies for @devcycle / @openfeature to patch versions

jonathannorris:
# fix: tighten package.json dependencies for @devcycle / @openfeature to patch versions

This change updates all SDK packages to use tilde (`~`) version ranges instead of caret (`^`) for internal `@devcycle` and `@openfeature` dependencies, restricting updates to patch versions only.

## Changes
- Updated 12 SDK package.json files across `sdk/` directory
- Changed dependency version ranges from `^` to `~` for all `@devcycle/*` packages
- Changed dependency version ranges from `^` to `~` for `@openfeature/*` packages  
- Preserved `^` ranges for `peerDependencies` and third-party packages

## Impact
- **Before**: `"@devcycle/types": "^1.30.0"` (allows minor + patch updates)
- **After**: `"@devcycle/types": "~1.30.0"` (allows patch updates only)

This ensures more predictable dependency resolution and reduces the risk of breaking changes from minor version updates in internal packages.

### DIFF
--- a/sdk/js-cloud-server/package.json
+++ b/sdk/js-cloud-server/package.json
@@ -19,7 +19,7 @@
     "license": "MIT",
     "homepage": "https://devcycle.com",
     "dependencies": {
-        "@devcycle/types": "^1.30.0",
+        "@devcycle/types": "~1.30.0",
         "cross-fetch": "^4.1.0",
         "fetch-retry": "^5.0.6",
         "lodash": "^4.17.21"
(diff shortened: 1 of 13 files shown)


jonathannorris:
> does our automated release process touch these? wondering if theres some automated bumping of our shared deps via lerna/nx

yea I looked at the release scripts, they should auto-increment these versions, but I didn't see anywhere they set `^`. I guess we will see on the next release what they do.

phaym (approved):

PR #1121, opened 2025-07-22
fix: add eval reasons to aggregate events

kaushalkapasi:
- fix: implements sending eval reasons as part of events metadata
- chore: update NodeJS eventQueue tests to verify eval metaData is properly defined

### DIFF
--- a/lib/shared/bucketing-assembly-script/assembly/eventQueue/requestPayloadManager.ts
+++ b/lib/shared/bucketing-assembly-script/assembly/eventQueue/requestPayloadManager.ts
@@ -6,6 +6,7 @@ import {
     FlushPayload,
     UserEventsBatchRecord,
     EventQueueOptions,
+    EVAL_REASONS,
 } from '../types'
 import { JSON } from '@devcycle/assemblyscript-json/assembly'
 import { _getPlatformData } from '../managers/platformDataManager'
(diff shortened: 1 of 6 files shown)


lib/shared/bucketing-assembly-script/assembly/eventQueue/requestPayloadManager.ts
```diff
@@ -92,23 +103,39 @@ export class RequestPayloadManager {
...
-                            null,
-                        )
-                        aggEvents.push(
-                            new DVCRequestEvent(
-                                dvcEvent,
-                                user_id,
-                                emptyFeatureVars,
-                            ),
-                        )
+                        const evalReasonAggMap = varAggMap.get('value')
```
phaym, line 106: why do we have these conditions checking for hardcoded 'value'?

phaym: oh for defaults? cus there would be no variation key?

kaushalkapasi: Yup, exactly. It'll always be `value` since there's no variation Id or any other unique identifier.

phaym (approved):
lgtm but we prob want someone more familiar to take a look

lib/shared/bucketing-assembly-script/assembly/eventQueue/requestPayloadManager.ts
```diff
@@ -25,6 +26,16 @@ export class RequestPayloadManager {
     private readonly chunkSize: i32
     private readonly clientUUID: string
 
+    private readonly evalReasonKeysToSkip: string[] = [
```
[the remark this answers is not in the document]
kaushalkapasi: nope, cleaning it up!

PR #1128, opened 2025-07-29
feat: add config metadata parsing through wasm

suthar26:
- adds configmetadata through wasm
- adding this to make sure there is no need for json parsing for sdk config in each language
- tested using hooks through the dev-apps for local and cloud

closes COR-3490


lib/shared/bucketing-assembly-script/assembly/managers/configDataManager.ts
```diff
@@ -17,3 +19,14 @@ export function _getConfigData(sdkKey: string): ConfigBody {
 export function _hasConfigData(sdkKey: string): bool {
     return _configData.has(sdkKey)
 }
+
+export function _getConfigMetadata(sdkKey: string): string {
+    if (_configMetadata.has(sdkKey)) {
```
JamieSinn, line 24: Check if we have this sdk key in the config map first.

JamieSinn (approved):

PR #1130, opened 2025-08-01
fix: update config metadata to stringify properly

suthar26:
- instead of stringifying the object, adding the project and environment to be objects and stringifying the parent object
- tested locally using devapp

### DIFF
--- a/sdk/nodejs/src/client.ts
+++ b/sdk/nodejs/src/client.ts
@@ -519,7 +519,6 @@ export class DevCycleClient<
         if (!this.bucketingLib) {
             return
         }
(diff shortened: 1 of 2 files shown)


sdk/nodejs/src/client.ts
```diff
@@ -519,6 +519,7 @@ export class DevCycleClient<
         if (!this.bucketingLib) {
             return
         }
+        console.warn(this.bucketingLib.getConfigMetadata(this.sdkKey))
```
JamieSinn, line 522: cleanup

PR #1131, opened 2025-08-05
fix(client): ReferenceError: Property 'document' doesn't exist on react native

santiagofm:
When using `@devcycle/react-client-sdk` we keep getting a reference error because the global `document` object doesn't exist on react-native.

Seems like we are guarding against this when [registering the visibility changed handlers](https://github.com/DevCycleHQ/js-sdks/blob/main/sdk/js/src/Client.ts#L916-L918), but not when closing the client and unregistering.

### DIFF
--- a/sdk/js/src/Client.ts
+++ b/sdk/js/src/Client.ts
@@ -715,7 +715,7 @@ export class DevCycleClient<
 
         this._closing = true
 
-        if (document && this.pageVisibilityHandler) {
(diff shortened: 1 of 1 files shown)


kaushalkapasi:
Hey @santiagofm thanks for opening this PR, I've made an internal PR with the [same change](https://github.com/DevCycleHQ/js-sdks/pull/1132) so that all the tests can run due to our security requirements. We should have this fix merged & released shortly!

PR #1137, opened 2025-08-11
feat: create dynatrace hook

jsalaber:


sdk/nodejs/src/hooks/DynatraceOtelLogHook.ts
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    onFinally<T extends DVCVariableValue>(
+        hookContext: HookContext<T>,
+        variableDetails: DVCVariable<T> | undefined,
+        variableMetadata: VariableMetadata | undefined,
+    ): void {
+        const span = this.spans.get(hookContext)
+
+        if (variableDetails) {
```
suthar26, line 53: add && span here so its one check for both?

jonathannorris:
@jsalaber / @suthar26 what's the status of this PR?

